PR #62, opened 2026-04-24
fix(tasks): guarantee terminal status tag when structured-result write throws

### LINKED ISSUE #57: Non-atomic task completion leaves stale `running` tag on throw
## Problem

The task completion path in `src/index.ts` is not atomic. If anything between the `result` write and the terminal status write throws, the task is left with the `running` tag set indefinitely despite having actually completed.

## Evidence

Task `tasks/20260424-061340-orchestrator-sweep` (codex, 2026-04-24):

- `result` entry: `created_at: 2026-04-24T06:30:05.393Z`, exit code 0 — the runtime finished successfully and the markdown result was written.
- `status` entry: `updated_at: 2026-04-24T06:29:53.327Z`, tags still contain `running`, `claimed_by:hugin-huginmunin-2599231`, `lease_expires:1777012313324` (06:31:53Z).

The status tag flip never happened, even though the runtime completed cleanly and the result was persisted.

## Root cause

In `src/index.ts`:

- Completion flow (around line 3150 → 3371): write `result` (markdown) → `writeStructuredTaskResult` (`result-structured`) → `munin.write(..., "status", ...)` to flip `running` → `completed`.
- The enclosing scope is `try {` at 2767 → `finally {` at 3436 — **no `catch`**.
- A throw in `writeStructuredTaskResult` (Zod validation, classification check) or in the status write itself bubbles up to pollLoop's generic `catch` at `src/index.ts:3496`, which just logs `Poll error:` and moves on. The task stays `running`.

Compounding this, the stale tag is not cleaned up promptly — see the companion issue for the reaper being blocked by long-running tasks.

## Suggested fix

Two options, either is fine:

1. **Status-first ordering.** Flip `status` to the terminal state *before* writing `result-structured`. The markdown `result` is the user-facing source of truth; `result-structured` is machine-readable and can be retried or reconstructed. Losing the structured result is recoverable; a stuck `running` tag is not.
2. **Guarded status write.** Wrap the `status` write in its own try/catch and a small retry loop. If `result` landed, the status flip must land too.

Either way, add a `catch` to the enclosing scope so the dispatcher at minimum knows a task completed with a post-processing error and can log something more actionable than `Poll error:`.

## Acceptance

- A task whose `writeStructuredTaskResult` throws still ends up with `completed`/`failed` tags.
- A unit test that injects a throw in the structured-result path asserts the status tag ends in a terminal state.

mvanhorn:
## Summary

Wraps `writeStructuredTaskResult` in try/catch in both completion paths of `src/index.ts` so a throw (Zod validation, classification check) no longer prevents the terminal status write. The lifecycle-critical `"status"` flip from `running` to `completed`/`failed`/`cancelled` always lands.

Fixes #57.

## Why this matters

The issue (self-filed) traces a stuck-running task back to `src/index.ts`: `writeStructuredTaskResult` runs before `munin.write(..., "status", buildTerminalStatusTags(...))`, both inside a `try` at 2767 whose enclosing scope has a `finally` at 3436 but **no** `catch`. A throw from the structured-result path propagates up to the `pollLoop` generic handler at 3496 and the task's `running` tag never flips. As the maintainer puts it: "Losing the structured result is recoverable; a stuck `running` tag is not."

## Changes

`src/index.ts`, both the cancelled branch (~line 3289) and the non-cancelled completion branch (~line 3338):

- Introduce `let postProcessingError: string | null = null;` at the top of each branch.
- Wrap `writeStructuredTaskResult(...)` in try/catch. On catch, record the error message, `console.warn` it, but do **not** await anything else in the catch — the next `await` is the terminal-status `munin.write`.
- After the terminal-status `munin.write`, if `postProcessingError` is set, best-effort `munin.log` the post-processing error inside its own try/catch so a degraded log call cannot surface as a new error.

The existing happy-path `munin.log` calls (`Task completed in ...s`, `Task cancelled in ...s`) stay where they were.

`munin.write(..., "status", ...)` itself is NOT wrapped — if it throws, the pollLoop's existing generic handler should still see it; we do not want to silently swallow lifecycle-critical failures.

### Ordering

The first revision of this patch awaited the diagnostic `munin.log` inside the catch, BEFORE the terminal status write. An internal review (`codex review`) pointed out that a degraded Munin would then re-introduce the exact stuck-running failure mode the patch is trying to prevent. The commit log reflects that second pass: status write first, diagnostic log best-effort after.

## Testing

```
npm run build   # tsc, passes
npm test        # 28 files, 400 tests, all pass
```

## Scope

Not included: the issue's "unit test that injects a throw in the structured-result path" acceptance bullet. The completion flow is currently in the ~3700-line `pollOnce` function inside `src/index.ts`; a testable unit would require extracting it. Happy to follow up with that refactor in a separate PR if helpful.

## AI disclosure

This contribution was developed with AI assistance (Codex).

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -3286,6 +3286,7 @@ async function pollOnce(): Promise<{ hadTask: boolean; queueDepth: number }> {
         undefined,
         taskClassification,
       );
+      let postProcessingError: string | null = null;
       try {
         await writeStructuredTaskResult(
           taskNs,
@@ -3310,18 +3311,10 @@ async function pollOnce(): Promise<{ hadTask: boolean; queueDepth: number }> {
           taskClassification,
         );
       } catch (err) {
-        const errMsg = err instanceof Error ? err.message : String(err);
+        postProcessingError = err instanceof Error ? err.message : String(err);
         console.warn(
-          `Failed to write structured result for ${taskNs}: ${errMsg}. Proceeding to terminal status write so the task does not stay 'running'.`
+          `Failed to write structured result for ${taskNs}: ${postProcessingError}. Proceeding to terminal status write so the task does not stay 'running'.`
         );
-        try {
-          await munin.log(
-            taskNs,
-            `Post-processing error: structured result write failed (${errMsg}). Terminal status still recorded.`
-          );
-        } catch {
-          // Don't let a log failure block the status write.
-        }
       }
       await munin.write(
         taskNs,
@@ -3331,11 +3324,22 @@ async function pollOnce(): Promise<{ hadTask: boolean; queueDepth: number }> {
         undefined,
         taskClassification,
       );
+      if (postProcessingError) {
+        try {
+          await munin.log(
+            taskNs,
+            `Post-processing error: structured result write failed (${postProcessingError}). Terminal status still recorded.`
+          );
+        } catch {
+          // Don't let a diagnostic log failure surface as a new error.
+        }
+      }
       await munin.log(
         taskNs,
         `Task cancelled in ${Math.round(durationMs / 1000)}s (reason: ${cancellation.reason}, executor: ${executorLabel})`
       );
     } else {
+      let postProcessingError: string | null = null;
       try {
         await writeStructuredTaskResult(
           taskNs,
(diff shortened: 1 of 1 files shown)


Magnus-Gille:
Closing as superseded. This PR (branch `osc/57-atomic-completion`) targeted #57, which was resolved on 2026-04-25 by a different, more robust implementation that is now on `main`: `finalizeTaskCompletion` in `src/task-helpers.ts` (see the `// --- Atomic task completion (#57) ---` section).

That helper provides the same guarantee this PR aimed for, with a strictly better ordering:
- **This PR:** write the structured result inside try/catch, *then* the terminal status.
- **main:** write the **terminal status FIRST** (guaranteed flip even if the structured write throws), *then* the structured result in try/catch, then the log. It also handles the single-owner CAS path for runtime-owned delivery (#68), which didn't exist when this PR was opened.

The helper is wired into both the cancelled and normal completion branches in `src/index.ts`. Issue #57 is already closed. No action needed here — the diff would conflict and regress the write ordering.

PR #64, opened 2026-04-26
orch-v1 Step 4: Pi-side broker endpoint

Magnus-Gille:
## Summary

Adds the orchestrator-v1 broker — a Tailscale-only HTTP service that accepts delegation envelopes from MCP clients, persists state in Munin, and exposes a per-task journal. Opt-in via `HUGIN_BROKER_KEYS` (when unset, the dispatcher runs unchanged).

Implements §3 (request envelope), §4 (await/result), §5 (journal events), §12.2/§12.3/§12.4/§12.5 (Munin durability + reconciliation) of `docs/orchestrator-v1-data-model.md`.

## Endpoints

| Method | Path | Behavior |
|--------|------|----------|
| POST | `/v1/delegate/submit` | Envelope validation, alias resolution, idempotency-key dedupe, Munin write, journal append, returns 202 |
| POST | `/v1/delegate/await` | Pure Munin read (§12.4) — terminal → result-structured/result-error; otherwise running |
| POST | `/v1/delegate/rate` | Appends `delegation_rated` event |
| GET/POST | `/v1/delegate/list` | Projects journal events to per-task rows |
| GET | `/v1/delegate/models` | Alias map + runtime rows + policy_version |

## What's wired

- Express app on its own port (default `:3033`, loopback bind; production sets `HUGIN_BROKER_HOST` to the Tailscale interface IP)
- Bearer-token middleware with constant-time compare; principals loaded from `HUGIN_BROKER_KEYS` (inline JSON) or `HUGIN_BROKER_KEYS_FILE`
- Append-only journal at `~/.hugin/delegation-events.jsonl`, forward-compat skip on unknown `event_schema_version`
- In-memory idempotency dedupe (`fresh` / `retry` / `collision`) with windowed expiry
- Two-phase complete (§12.3): write `result-structured`, then CAS the status entry's lifecycle tag
- `orch-v1` tag on all broker submissions to keep them out of the legacy poll loop
- Reconciliation sweep (default 60s) that backfills `delegation_submitted` events for tasks visible in Munin but missing from the journal — idempotent
- Startup/shutdown wired into `src/index.ts`; `runningBroker?.close()` and `brokerReconciler?.stop()` integrated into the main shutdown path

## What's not yet wired (deferred to Step 5/5b)

- Reconciliation cases (3) and (4) — crash-between-writes recovery and lease-expired-without-result — need lease metadata that lands with the executor in Step 5/5b. The reconciler framework is in place but those handlers only fire when leases exist.
- No executor consuming `pending` orch-v1 tasks yet; submissions land in Munin and stay `pending` until Step 5 arrives.

## Test plan

- [x] `npm run build` — clean
- [x] `npm test` — 501 tests pass (62 new for broker)
- [x] Unit coverage: auth (12), idempotency (7), journal (7), task-store (9), alias-resolution (5), reconciliation (4)
- [x] Integration coverage: handlers (18) — exercises the live Express app on `port: 0` via Node's built-in `fetch`, mocking only the `MuninClient`
- [ ] Cross-model adversarial review (`/review-pr-codex`)
- [ ] Manual smoke on Pi (deferred until Step 5 executor exists — no executor means no end-to-end happy path to test yet)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -126,7 +126,17 @@ hugin/
 │   ├── task-result-schema.ts     # Structured task result with classification
 │   ├── task-status-tags.ts       # Tag manipulation helpers for task lifecycle
 │   ├── task-graph.ts             # Task dependency graph for pipelines
-│   └── result-format.ts          # Result formatting utilities
+│   ├── result-format.ts          # Result formatting utilities
+│   └── broker/                   # Orchestrator-v1 broker (Tailscale-only HTTP, /v1/delegate/*)
+│       ├── server.ts             # Express app + opt-in startup (HUGIN_BROKER_KEYS)
+│       ├── handlers.ts           # submit/await/rate/list/models endpoint handlers
+│       ├── auth.ts               # Bearer-token middleware, constant-time compare
+│       ├── idempotency.ts        # In-memory idempotency-key dedupe (§3.1)
+│       ├── journal.ts            # Append-only delegation-events.jsonl + projection
+│       ├── task-store.ts         # Munin operations: submit / read / two-phase complete
+│       ├── alias-resolution.ts   # Alias → AliasResolved annotation + policy_version
+│       ├── reconciliation.ts     # Periodic sweep: backfill journal events for orch-v1 tasks
+│       └── types.ts              # Zod schemas for the wire contract
 ├── tests/                        # 19 test files mirroring src/
 ├── docs/                         # Engineering plans, evaluations, security docs
 │   └── security/                 # Threat models and security assessments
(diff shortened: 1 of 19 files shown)


Magnus-Gille:
## Codex review response

Fixes for F2–F6 pushed in 08a1cf8. Quick map of decisions:

- **F3 (idempotency race)** — Fixed. `IdempotencyIndex.inspect()` replaced with atomic `reserve()` that returns `in_flight` when a concurrent submission holds the same key. Submit handler returns `503 in_flight` and calls `release()` on Munin write failure so retries proceed.
- **F4 (canonicalize drops nested fields)** — Fixed. `JSON.stringify(req, Object.keys(req).sort())` swapped for a recursive canonicalizer. Added tests covering `worktree.target_files` and key-order stability.
- **F5 (unbound submitter)** — Fixed. Submit handler rejects with 400 when `orchestrator_submitter !== principal`.
- **F2 (`/await` lease/stale)** — Partially fixed. Dropped `max_wait_s` from the schema (not implemented). The empty-lease + `running` response stays for now: there is no executor yet to populate lease metadata, so `stale` cannot meaningfully be computed in this PR. Comment in the handler points at Step 5 where lease/stale logic lands alongside the executor.
- **F6 (doc drift)** — Fixed. Spec now says `/v1/delegate/submit`, `/v1/delegate/models`, and `HUGIN_BROKER_RECONCILIATION_INTERVAL_MS`.

### Push-back on F1 (broker accepts work nothing executes)

Intentional. This PR is the broker half of a phased rollout: the orch-v1 executor lands in Step 5 (#5). Submitted tasks sit in `pending` with the `orch-v1` tag and are deliberately filtered out of the legacy poll loop (`src/index.ts:2593-2600`) so they don't get picked up by the wrong runtime. The path is exercised end-to-end by the MCP work in Step 6 once the executor exists. Disabling submit until then would force a flag-flip plus broker restart at Step 5 with no test coverage in between.

The broker is opt-in (`HUGIN_BROKER_KEYS` must be set), and in production the only client is the not-yet-shipped MCP, so the "accepts work it can't execute" risk is bounded to the dev loop.

All 506 tests pass.

PR #71, opened 2026-05-17
fix: harden update-cli.sh — cross-major bumps, failure logging, smoke test (#60)

### LINKED ISSUE #60: `update-cli.sh`: missed-major bumps, silent failures, no smoke test
## Problem

`scripts/update-cli.sh` runs daily via cron on the Pi and is the only thing keeping `@anthropic-ai/claude-code` and `@openai/codex` current. Three weak spots:

1. **Missed major bumps.** `npm update -g "$pkg"` respects the stored semver range. When installed with `npm i -g <pkg>`, npm pins a caret range, so `npm update` will move within the major but silently skip major releases. A 1.x → 2.x of codex would never land on the Pi and there would be no signal.
2. **Silent failures.** The script only `logger`s on a version *change*. If `npm update` errors out (network blip, registry 5xx, disk full), cron captures stdout but nothing reaches syslog and nothing reaches Munin. The Pi has no mail reader, so cron MAILTO output also goes nowhere.
3. **No post-update smoke test.** An update that leaves `claude --version` or `codex --version` broken would pass silently. A trivial `--version` probe after the update would catch install corruption.

Also: bubblewrap (see #59) is not covered by this script — it's an apt package, not npm. Out of scope for this issue but worth noting that system-level codex dependencies need a separate drift check.

## Suggested fix

Small edits to `scripts/update-cli.sh`:

- Replace `npm update -g "$pkg"` with `npm i -g "${pkg}@latest"` to force cross-major upgrades.
- Wrap the install in an exit-code check; `logger -t hugin-update "FAILED: $pkg install returned non-zero"` on failure, don't swallow.
- After both updates, run `claude --version` and `codex --version` and `logger` the results (and fail the script if either is missing).
- Consider writing a one-line status to Munin (`meta/hugin/cli-versions` or similar) so the other Grimnir environments can see CLI drift without SSH'ing to the Pi.

## Acceptance

- A broken codex install (e.g. truncated tarball) produces a `hugin-update` syslog entry flagged as failure.
- A major version bump is actually applied on the next cron run.
- A one-line Munin entry reflects the currently-installed versions.

Magnus-Gille:
## Summary

Addresses all three weak spots in `scripts/update-cli.sh` from #60:

1. **Missed major bumps** — `npm update -g` → `npm i -g <pkg>@latest`, forcing cross-major upgrades.
2. **Silent failures** — install wrapped in exit-code check; `logger -t hugin-update 'FAILED: ...'` on non-zero; script exits 1.
3. **No smoke test** — `claude --version` / `codex --version` run after update; a broken install fails the script + emits a syslog entry.

Bonus: writes a one-line status to Munin (`meta/hugin/cli-versions`) so other Grimnir environments see CLI drift without SSH. Best-effort — Munin/network failure never fails the update.

## Implementation notes

- `smoke_test` prints the version to stdout, all human/log lines to stderr, so cron output is preserved while the caller captures only the version.
- `FAILED` is set in the parent shell (command substitution runs in a subshell, so the sentinel `broken` is detected by the caller).
- `.env` sourced for `MUNIN_API_KEY`; absent key just skips the Munin write.

## Verification

- `bash -n` syntax check: clean
- Helper logic unit-tested locally (stdout/stderr separation, broken-sentinel → FAILED=1)
- Full acceptance (cron syslog entry on broken install, major bump applied, Munin entry) requires a Pi cron run — code paths verified by inspection + local harness

Closes #60

### DIFF
--- a/scripts/update-cli.sh
+++ b/scripts/update-cli.sh
@@ -3,31 +3,51 @@ set -euo pipefail
 
 # Update Claude Code and Codex CLI tools
 # Intended to run daily via cron on the Pi
-# Logs version changes to syslog
+# Logs version changes and failures to syslog (tag: hugin-update)
+# Writes a one-line version status to Munin (meta/hugin/cli-versions)
 
 export PATH="/usr/local/bin:/usr/bin:/bin:$HOME/.npm-global/bin:$PATH"
 
+# Source the Pi .env for MUNIN_API_KEY (best-effort; Munin write is non-fatal)
+if [ -f "$HOME/hugin/.env" ]; then
+  set -a
+  # shellcheck disable=SC1091
+  . "$HOME/hugin/.env"
+  set +a
+fi
+
+MUNIN_URL="${MUNIN_URL:-http://localhost:3030}"
+
+# Tracks whether any package install failed (script exits non-zero at the end)
+FAILED=0
+
+installed_version() {
+  local pkg="$1"
+  npm ls -g "$pkg" --depth=0 --json 2>/dev/null | node -e "
+    let d=''; process.stdin.on('data',c=>d+=c); process.stdin.on('end',()=>{
+      try { const j=JSON.parse(d); console.log(j.dependencies?.['${pkg}']?.version || 'not-installed'); }
+      catch { console.log('not-installed'); }
+    });
+  "
+}
+
 update_package() {
   local pkg="$1"
-  local old_version
+  local old_version new_version
 
-  old_version=$(npm ls -g "$pkg" --depth=0 --json 2>/dev/null | node -e "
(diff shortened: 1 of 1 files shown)


Magnus-Gille:
Codex cross-model review (gpt-5.4, xhigh) — 3 rounds:

- Round 1: 2 medium + 2 low findings (installed_version abort under set -e; Munin false-success; smoke vs installed version; token in argv)
- Round 2: all 4 resolved; 1 new low (`--fail-with-body` curl ≥7.76 dependency)
- Round 3: **No findings.** Clean.

`bash -n` clean; failure paths exercised locally. Ready for human merge.

Issue #78, opened 2026-05-29
Skill distillation A1: land #77 crash-recovery liveness + kill-during-local-execution test


Magnus-Gille:
**Gating step for the whole skill-distillation lane** (Codex's single most important next step).

Fix Hugin #77 (crash-recovery liveness): a delivery crash + auto-restart must auto-reconcile rather than waiting for a 2nd post-lease-expiry restart. Then prove it with a **kill-during-local-execution acceptance test wired through the same task lifecycle the skill lane will use**.

Why this gates skill distillation specifically: the local lane accepts 10min–3h batch runs. A liveness bug there strands a route decision non-terminal, delays fallback, and corrupts promotion telemetry. Requires: stable (host-based) worker identity, lease-expiry reconciliation, startup recovery, terminal-state correctness, idempotent artifact delivery.

Scope note: gates `active` routing and any Hugin-integrated shadow evidence — does NOT block offline package/eval/profile authoring (A3/A6).

Source: cross-model debate (Codex gpt-5.5, 2 rounds), 2026-05-29. Artifacts in https://github.com/Magnus-Gille/hugin/tree/main/docs/decisions/skill-distillation/ (summary.md, critique-log.json).

Magnus-Gille:
## Verified on the Pi (2026-05-30)

PR #86 (#77 fix) deployed to huginmunin. Health confirms the host-stable worker identity is live: `worker_id: hugin-huginmunin` (no PID suffix), `process_instance_id: hugin-huginmunin-1060195`.

**Crash-recovery acceptance test (deterministic, through the real task lifecycle):** fabricated the exact post-crash deadlock state in Munin — a `running + delivery:pending` checkpoint `claimed_by:hugin-huginmunin` with a **live** lease (now+100s), a valid `### Artifacts` manifest, and a staged local file — then restarted Hugin once.

Result: on the **first** restart, `recoverStaleTasks` recognised the orphan as ours (`isOurs` via the host-stable id), reconciled it, rsync+verified the artifact to the NAS (`/home/magnus/mimir-inbox/e2e77.md`, content match), and flipped the task to terminal `["completed", "delivery:verified"]` — the `delivery:pending` / `claimed_by` / `lease_expires` tags reclaimed and cleared.

[...]

Issue #77, opened 2026-05-18
Crash mid-delivery: orphaned delivery:pending never auto-reconciles (PID-based workerId + reaper defers to one-shot startup scan)


Magnus-Gille:
## Summary

The Pi e2e for #75 (kill Hugin mid-delivery → restart) found a **liveness bug** in the runtime-owned artefact-delivery crash-recovery path (#68). The #68 *safety* guarantee holds (agent content durably checkpointed, agent runs exactly once, no silent data loss, no paid rerun), but a delivery interrupted by a real process crash is **not automatically reconciled**: the task stays `running` + `delivery:pending` forever and never reaches a terminal state until a *second*, manual Hugin restart that happens after the dead worker's lease has expired.

## Mechanism (deadlock between two components)

1. `workerId` is **PID-derived**: `hugin-<host>-<pid>` (see `/health` `worker_id`). After a `kill -9` + `systemd Restart=always`, the new process has a new PID → a new `workerId`.
2. `recoverStaleTasks()` runs **only once at startup** (`src/index.ts:1882` comment) and is gated by the single-owner test at `src/index.ts:1799-1811`:
   - `isOurs = claimedBy === workerId || claimedBy === null` → **false** (task is `claimed_by:hugin-<host>-<oldpid>`, restarted process has a new id).
   - The dead worker's lease (`lease_expires:*`) is still live at restart (RestartSec=10s « lease ~120s) → `leaseExpired` false.
   - `!isOurs && !leaseExpired` → `continue` at line 1804, **before** the `delivery:pending` reconcile branch at `src/index.ts:1821`. The reconcile branch is unreachable for a true crash-restart.
3. The lease reaper (`reapExpiredLeases`, periodic) **deliberately skips** `delivery:pending` (`src/index.ts:1926-1933`, `"left for startup reconciliation"`) — deferring to a startup scan that never re-runs.

Net: startup recovery defers to the reaper's window; the reaper defers back to startup recovery; neither ever reconciles. Orphaned indefinitely.

## e2e evidence (2026-05-18, Pi, Hugin @ 3fc15d3 + #74 fixes)

Watcher SIGKILLed Hugin during the live `rsync` of a 24 MB artefact (agent had completed once, cost `$0.0526`).

- Checkpoint correct & durable: `running, delivery:pending, claimed_by:hugin-huginmunin-221281, lease_expires:…`; agent `### Response` preserved in `result`.
[...]

Issue #75, opened 2026-05-18
Verify runtime-owned delivery (#68): Pi e2e for crash-recovery & cancel-race


Magnus-Gille:
Follow-up to #68 / PR #74 (merged, commit 2e2d083).

The runtime-owned artefact-delivery implementation is on `main` with 648 unit tests green, but the **crash-recovery** and **cancel-race** reconciliation paths in `src/index.ts` have no unit test seam (code-review findings A & C). The logic mirrors the tested active path; the reconciliation behavior itself is unverified.

### Mandatory deploy ordering
New `research-spike` SKILL.md against old Hugin reintroduces #68's silent data loss. Sequence:

1. `./scripts/deploy-pi.sh` — gate enabling `HUGIN_DELIVERY_POLICY=require` on the deploy preflight passing.
2. **Pi e2e:**
   - tiny manifest spike → confirm via `~/.hugin/logs/<task>.log` that **Hugin** (not the agent) ran rsync; files land in `mimir-inbox`; `result` has runtime-authored `### Artifact Delivery`.
   - deliberately bad `remote` → task `failed`, `Exit code: 2`, content preserved.
   - kill Hugin mid-delivery → restart reconciles (no paid rerun, correct terminal state, Ratatoskr reports correctly).
   - cancel during delivery → `cancelled` with `delivery:*` markers preserved.
3. Only then update `~/.claude/skills/research-spike/SKILL.md`.

Closes the verification gap noted in https://github.com/Magnus-Gille/hugin/issues/68#issuecomment-4476576237.

Magnus-Gille:
## Pi e2e results (2026-05-18, Hugin @ 3fc15d3 = #74 fixes + #76 ProtectSystem fix, deployed)

Infra blocker resolved first: `ProtectSystem` at *any* value (strict/full/strict) read-only-bind-mounts `/usr`, tripping OpenSSH 10's strict config-ownership check on the symlinked `20-systemd-ssh-proxy.conf` and breaking the NAS rsync/ssh. Fix = drop `ProtectSystem` entirely, keep the other hardening (#76, merged + deployed).

| Scenario | Result |
|---|---|
| **S1** happy-path delivery | ✅ PASS — file on NAS, sha256 matches result body, no `.partial`, terminal `completed`+`delivery:verified`, runtime-authored `### Artifact Delivery` |
[...]

PR #86, opened 2026-05-29
fix: positive failure exit codes (#73) + crash-recovery liveness (#77)

### LINKED ISSUE #73: Bug: recovery/reaper/shutdown write `Exit code: -1` which Ratatoskr mis-renders as success
Discovered during #68 mitigation #3.

Ratatoskr decides task success by matching the human `result` markdown against `/\*\*Exit code:\*\*\s*(\d+)/` and treats a **non-match as success**. A negative code fails `(\d+)` → **no match → mis-rendered as success**.

Three pre-existing dispatcher failure paths write `- **Exit code:** -1`:

- `recoverStaleTasks` (stale-task recovery) — `src/index.ts` (`Task recovered (...)`)
- `reapExpiredLeases` (lease reaper) — `src/index.ts` (`Lease expired ...`)
- `shutdown` (dispatcher shutdown) — `src/index.ts` (`Task interrupted by dispatcher shutdown ...`)

A task failed by any of these is reported to the user as **success**. The #68 work uses a positive `Exit code: 2` for its new delivery-failure path and the new `reconcileDeliveryPending` path, but did **not** change these three legacy paths (out of scope for #68 — filed separately as agreed in the Codex debate).

**Fix:** change these three sites to a positive integer (e.g. `1` or `2`) and add a regression test asserting Ratatoskr's regex reports failure for each. Audit `createFailureStructuredResult` default `exitCode: options.exitCode || -1` similarly.

Magnus-Gille:
Fixes #73 and #77 — two coupled defects in the dispatcher's failure/recovery subsystem.

## #73 — negative exit codes mis-rendered as success
Ratatoskr decides task success by matching the `result` markdown against `/\*\*Exit code:\*\*\s*(\d+)/` and treats a **non-match as success**. Every dispatcher-side failure path emitted `- **Exit code:** -1`, which fails `(\d+)` → no match → a **failed** task was reported as a **success**.

- New `DISPATCHER_FAILURE_EXIT_CODE = 1` constant, used at all 8 failure sites (recovery, reaper, shutdown, security rejection, approval rejection, generic failure ×3) and the `createFailureStructuredResult` default.
- Regression test `tests/exit-code-failure.test.ts` exercises the exact Ratatoskr regex + a source-scan guard asserting no failure path emits a negative `**Exit code:**`.

## #77 — crash mid-delivery never auto-reconciles
A `kill -9` + systemd restart during artefact delivery left a `running + delivery:pending` checkpoint that **never reached a terminal state** until a *second*, post-lease-expiry restart:
- `workerId` was PID-derived, so the restarted process saw the dead incarnation's task as "not ours";
- while the dead worker's lease was still live, the one-shot startup scan **skipped** it;
- the lease reaper **deliberately deferred** `delivery:pending` back to the startup scan → deadlock.

**Fix (defence in depth):**
1. **Host-stable `workerId`** (`hugin-<host>`, no PID). systemd runs exactly one Hugin per host, so a restarted process re-adopts and reconciles its own in-flight tasks immediately. The PID is retained as `process_instance_id` for observability (`/health`, heartbeat, startup log).
2. **The reaper now reconciles `delivery:pending` itself** once the lease has expired (owning worker provably dead) instead of deferring. `reconcileDeliveryPending` takes a `MuninClient` param so the reaper routes through `reaperMunin`; its CAS reclaim makes startup-scan-vs-reaper mutually safe (no double-delivery).
3. A **live in-process delivery stays protected** — the reaper never touches the currently-executing task (`shouldReapExpiredLease`'s `currentTask` guard), even though lease renewal stops before delivery.

The startup-recovery gate is extracted into a pure `decideStartupRecovery()` in `task-helpers.ts` (mirroring `shouldReapExpiredLease`), with regression tests in `tests/startup-recovery.test.ts` covering the exact #77 scenario (own dead incarnation, live lease, delivery:pending → reconcile, not skip).

## Testing
- `npm test`: 648 → **660 green** (12 new tests).
- `tsc --noEmit` clean.

## Not in this PR
- The **Pi e2e re-run** of #75 S3 (kill mid-delivery → restart auto-reconciles) — needs Pi hardware access. This PR makes the fix unit-test-deterministic; the e2e remains tracked in #75 / #78.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -329,6 +329,17 @@ let currentOllamaAbort: AbortController | null = null;
 // Runtime-owned artefact delivery (issue #68). Aborted by operator cancel /
 // shutdown so a hung `ssh`/`rsync` cannot wedge the single dispatcher slot.
 let currentDeliveryAbort: AbortController | null = null;
+// Separate abort slot for RECONCILE-path deliveries (`reconcileDeliveryPending`,
+// driven by startup recovery and the lease reaper). The reaper runs on its own
+// timer concurrently with the poll loop, so its reconcile must NOT share
+// `currentDeliveryAbort` with the live in-process delivery — clobbering that
+// shared slot would leave the live delivery un-abortable by operator cancel /
+// shutdown (review F1, #77). At most one reconcile runs at a time (startup
+// completes before the reaper timer arms; the reaper processes tasks serially
+// under `leaseReaperInFlight`), so a single slot is sufficient. Shutdown aborts
+// both slots; operator cancel targets only the live delivery (a reconcile is
+// always for a non-current, dead-owner task).
+let currentReconcileAbort: AbortController | null = null;
 let server: Server;
 let runningBroker: RunningBroker | null = null;
 let brokerReconciler: BrokerReconciler | null = null;
@@ -1670,7 +1681,9 @@ async function reconcileDeliveryPending(
   } else {
     const logPath = path.join(LOG_DIR, `${extractTaskId(taskNs)}.log`);
     const abort = new AbortController();
-    currentDeliveryAbort = abort;
+    // Reconcile-path slot, NOT the live-delivery slot (review F1): a reaper-driven
+    // reconcile must not clobber a concurrent live delivery's abort controller.
+    currentReconcileAbort = abort;
     try {
       delivery = await deliverArtifacts({
         manifest: task.artifactManifest,
@@ -1690,7 +1703,7 @@ async function reconcileDeliveryPending(
         signal: abort.signal,
       });
     } finally {
-      currentDeliveryAbort = null;
+      currentReconcileAbort = null;
     }
     if (!delivery.ok) {
       // missing-local / unsafe-local (no trustworthy deliverable) are ALWAYS
@@ -4247,6 +4260,11 @@ async function shutdown(signal: string): Promise<void> {
   if (currentDeliveryAbort) {
     console.log("Aborting running artefact delivery...");
     currentDeliveryAbort.abort();
+  }
+
+  if (currentReconcileAbort) {
+    console.log("Aborting running delivery reconciliation...");
+    currentReconcileAbort.abort();
   }
 
   if (currentSdkAbort) {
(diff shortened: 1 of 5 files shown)


Magnus-Gille:
## Adversarial review (in-harness opus reviewer) — SHIP

Reviewed the full diff + surrounding code (`reconcileDeliveryPending`, `recoverStaleTasks`, `reapExpiredLeases`, the delivery checkpoint, `shutdown`, `killOrphanDispatchers`) and reconstructed the original inline gate from main.

**Verdict:** #73 and #77 are fixed correctly. The four high-risk concerns all check out:
- **Double-delivery / double terminal flip** — handled: both startup scan and reaper CAS-reclaim first; the loser bails before any rsync.
- **Live in-process delivery vs reaper** — handled: `currentTask` is set for the entire delivery window (incl. the cancelled-finalize CAS-lost path), and the reaper's `shouldReapExpiredLease` returns `reap:false` for `namespace === currentTask`.
- **Exit codes** — handled: every failure site emits the positive constant; no `-1` remains; cancelled stays `"CANCELLED"` (correct — operator action, not dispatcher failure).
- **`decideStartupRecovery` extraction** — verified faithful; the dropped `legacyStale` branch was provably unreachable dead code.

**One finding fixed in-branch (F1, MED):** routing the reaper through `reconcileDeliveryPending` made it clobber the shared `currentDeliveryAbort` global, which could leave a *concurrent* live delivery un-abortable by operator cancel / shutdown (no corruption / no double-delivery). Fixed in `ea7b06f`: the reconcile path now uses a separate `currentReconcileAbort` slot; shutdown aborts both. Added `tests/delivery-abort-isolation.test.ts`.

**Note (F4, LOW — operational, not a code defect):** the host-stable `workerId` assumes systemd runs exactly one Hugin per host. `killOrphanDispatchers` enforces that on the happy path. Manually co-running `npm run dev` alongside the systemd unit in the same repo dir is unsupported (both would share `hugin-<host>`); `process_instance_id` is carried for observability but intentionally not used in reclaim decisions.

Issue #16, opened 2026-04-05
Missing dependencies block tasks forever


Magnus-Gille:
## Bug

In `task-graph.ts`, the `evaluateBlockedTask` function treats dependencies with state `missing` (i.e., the task entry doesn't exist in Munin at all) as neither failed nor completed. This means:

- `shouldPromote` is false (not all completed)
- `shouldFail` is false (no failed deps when policy is `fail`, only missing ones)

A task with a typo in its `depends-on:` tag will block indefinitely with no timeout or escalation.

## Found by

AI user testing review (Claude Opus 4.6 code review, 2026-04-05). See `docs/ai-user-testing-review.md`.

## Suggested fix

Treat `missing` dependencies the same as `failed` when evaluating `shouldFail` (when policy is `fail`). A dependency that doesn't exist is not going to appear — it's a permanent failure, not a transient wait.

Update `evaluateBlockedTask` in `src/task-graph.ts` so that `missingIds` count toward the failure condition. Add test cases for the missing-dependency scenario.

## Files

- `src/task-graph.ts` (primary)
- `tests/task-graph.test.ts` (add tests)

## Acceptance criteria

- A blocked task with a `depends-on:nonexistent-task` tag should transition to `failed` (when policy is `fail`) instead of blocking forever
- Existing tests continue to pass
- New test covers the missing-dependency case

Issue #14, opened 2026-04-04
False positive: task classified as client-confidential due to namespace references in prompt


Magnus-Gille:
## Bug

Tasks that mention `clients/*` namespace patterns in their prompt (e.g., code that handles client namespaces) get classified as `client-confidential` sensitivity, which blocks execution with:

```
Runtime "claude" cannot execute private-sensitivity work (max allowed: internal)
```

### Reproduction

Submit a task whose prompt contains references to `clients/*` namespaces as part of technical instructions (e.g., "consolidate `projects/*` and `clients/*` namespaces"). The Librarian classifies the entire task as `client-confidential` even though the prompt contains no actual client data — only namespace pattern references in code/specs.

### Example

Task `20260404-200728-consolidation-sprint3` — a code implementation task for munin-memory that mentions `clients/*` as a namespace pattern in the spec. Contains zero client data. Blocked by sensitivity classification.

### Expected behavior

Classification should distinguish between:
- Referencing namespace patterns in code/specs (`clients/*` as a string in instructions) → `internal`
- Actual client data (names, details, confidential content) → `client-confidential`

### Workaround

Unclear — either the task prompt needs to avoid mentioning `clients/*`, or the classification ceiling for the task namespace needs adjusting, or the runtime sensitivity cap needs raising.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #58, opened 2026-04-24
`reapExpiredLeases` cannot run while dispatcher is busy on a long task


Magnus-Gille:
## Problem

`reapExpiredLeases` is called inside the poll loop (`src/index.ts:3480`, every 5 polls), but `pollOnce` blocks on the currently-executing task. If a task has a long timeout (e.g. research spikes with `Timeout: 5400000` = 90 min), the reaper cannot run for up to that duration, even though it only touches *other* tasks' stale tags.

## Evidence

On 2026-04-24:

- `tasks/20260424-061340-orchestrator-sweep` (codex) completed at 06:30:05Z but kept its `running` tag due to #57.
- `tasks/20260424-062000-drone-c2-protocol-landscape` (claude, 90-min timeout) was claimed at 06:21:32Z and was still running when this was observed at ~08:32 CEST.
- Dashboard showed two tasks as `RUNNING` simultaneously. The codex one was orphaned; the reaper could not clean it up because `pollOnce` was blocked on the drone task.

In practice, a stale `running` tag survives as long as the next task takes to finish (or until the dispatcher restarts and `recoverStaleTasks` runs).

## Suggested fix

Run `reapExpiredLeases` on its own independent `setInterval` timer, like `startLeaseRenewal` (`src/index.ts:1398`) and `startCancellationWatch` already do. The reaper only acts on tasks not owned by this worker (or tasks whose lease has expired), so it's safe to run concurrently with a task.

Cadence: every ~60s is fine. Keep the poll-loop invocation too (cheap) or drop it — preference for one source of truth via the independent timer.

## Acceptance

- With the dispatcher running a 30-min task, an orphaned `running` task submitted immediately after gets reaped within ~1 minute.
- Reaper does not interfere with normal lease renewal for the current task (guarded by `claimed_by` / ownership check already in place).

PR #109, opened 2026-06-17
feat(homeserver): forward modelId/frontierModelId/verify on /delegate (#14)

Magnus-Gille:
The hugin half of home-server-inference-evaluation #14 (pairs with its PR #16).

Now that the gateway `/delegate` handler accepts `modelId`, `frontierModelId`, and an opt-in `verify` spec, re-enable the executor to send them (they were stripped during the #107 review because the gateway ignored them):
- adds `HomeserverVerifySpec` to the config
- forwards `task.model → modelId`, `task.frontierModelId`, and `task.verify` on the delegate path
- updates the stale "gateway ignores these" comments
- test: asserts the fields are sent + a new `verify`-passthrough case (17 tests, typecheck clean)

This activates the executor's dormant escalation and lets nightly delegations carry a deterministic check so the ledger learns real pass/fail.

### DIFF
--- a/src/homeserver-executor.ts
+++ b/src/homeserver-executor.ts
@@ -26,18 +26,30 @@ import * as path from "node:path";
 
 export type HomeserverPath = "chat" | "delegate";
 
+/** Declarative grading forwarded to the gateway /delegate `verify` field (gateway-api-contract.md). */
+export interface HomeserverVerifySpec {
+  nonEmpty?: boolean;
+  answerIs?: string;
+  contains?: string[];
+  matches?: string;
+  numeric?: number;
+  json?: boolean;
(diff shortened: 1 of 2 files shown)


Magnus-Gille:
Closing as stale. This forwards `modelId`/`frontierModelId`/`verify` matching the inline `verify` spec contract from PR #16 (home-server-inference-evaluation), but that PR was closed as superseded by #58, which reimplemented issue #14 against current main with a different shape — a `verifier-registry.ts` allow-list spec instead of the inline spec this PR assumes.

Re-enabling the executor's escalation/verify passthrough needs to be redone against #58's actual contract.

PR #125, opened 2026-07-02
feat(orchestrator): configurable worker max_tokens + surface finish_reason:length (#112)

Magnus-Gille:
Closes #112.

## Problem

`DirectModelExecutor` hard-capped every orchestrator call at `max_tokens: 4096` with no task/env/role override, and `extractChatCompletion` ignored `finish_reason`. A length-truncated response was returned as `ok:true`, so long worker/synthesis output could be **silently incomplete while the task reported success**.

## Fix

**1. Configurable completion-token cap** (4096 safe default)
- `HUGIN_ORCH_MAX_TOKENS` env → `OrchestratorConfig.maxTokens` → invoker defaults → `WorkerRequest.maxTokens` → request body.
- Per-role override via optional `RoleBinding.maxTokens` (e.g. a synthesizer that needs longer output than a worker).

**2. Parse `finish_reason`**
- `extractChatCompletion` now returns `finishReason`; `WorkerResult.truncated` is set when it is `"length"`.

**3. Surface truncation instead of reporting clean success**
- The engine collects a warning per truncated planner / worker / synthesizer response into `OrchestrationResult.warnings`.
- The executor renders a `### Warnings` section in the result summary and logs each warning.

## Tests

Red/green TDD — 14 new tests, confirmed failing before implementation:
- `worker-executor`: default cap = 4096, `req.maxTokens` override, `truncated` true/false for `finish_reason` = length/stop/absent.
- `model-invoker`: `defaults.maxTokens` and per-role `RoleBinding.maxTokens` threading.
- `engine`: warnings for truncated planner/worker/synth; empty array when clean.
- `config`: `HUGIN_ORCH_MAX_TOKENS` default/override/bad-int.
- `orchestrator-executor`: `### Warnings` section rendered on truncation.

Full suite green: **1151 tests pass**, build clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #126, opened 2026-07-02
feat(orchestrator): thread AbortSignal so timeout/cancel aborts in-flight model calls (#110)

### LINKED ISSUE #110: orchestrator: thread AbortSignal so task timeout/cancel aborts in-flight model calls
From the Codex review of PR #108 (finding #1, medium).

`runOrchestratorTask` races `runOrchestration()` against a task-level timeout, but the losing engine promise is never cancelled. `DirectModelExecutor` builds its own AbortController tied only to the per-call timeout, and `WorkerRequest`/`ModelInvoker.invoke()` have no signal field. Shutdown/cancel aborts SDK + Ollama work but not `currentOrchestratorAbort`, and the dispatcher `finally` doesn't clear that slot. A timeout/cancel can finalize the task while planner/worker/synth calls keep spending until their per-call timeout.

**Fix:** thread `AbortSignal` through `ModelInvoker.invoke()` + `WorkerRequest`; combine caller signal with per-call timeout in `DirectModelExecutor`; kill `PiHarnessExecutor` children on abort; abort the orchestrator controller when the task timeout wins; clear `currentOrchestratorAbort` in shutdown/finally. Add tests proving cancel aborts fetch/child work and blocks later calls.

Mitigation meanwhile: per-call timeouts bound spend.

Magnus-Gille:
Closes #110.

> **Stacked on #125** (base = `fix/orch-max-tokens-finish-reason`). GitHub retargets this to `main` once #125 merges. Review #125 first.

## Problem

`runOrchestratorTask` raced `runOrchestration()` against a task-level timeout but never cancelled the losing engine promise. `DirectModelExecutor` built its own `AbortController` tied only to the per-call timeout, and neither `WorkerRequest` nor `ModelInvoker.invoke()` carried a signal. So a timeout/cancel could finalize the task while planner/worker/synth calls **kept spending until their per-call timeout**. Shutdown/cancel aborted `currentOrchestratorAbort`, but the signal reached nothing, and the `pollOnce` `finally` never cleared that slot.

Current behaviour was *safe* (per-call timeouts bound spend) — this tightens cancellation so a timeout/cancel stops spend promptly.

## Fix

- `WorkerRequest.signal` + `ModelInvoker.invoke(opts.signal)`; `createModelInvoker` threads it into every `WorkerRequest`.
- `runOrchestration` accepts `opts.signal` and passes it to **every** invoke (planner, workers, verifier, synthesizer).
- **`DirectModelExecutor`**: short-circuits an already-aborted call before spending; combines `req.signal` with the per-call timeout controller so an external abort cancels the in-flight `fetch`; reports timeout vs cancel distinctly; removes the listener in `finally`.
- **`PiHarnessExecutor`**: short-circuits when pre-aborted; kills the child when the signal fires mid-run (`abortedBySignal` distinguishes it from a timeout kill).
- **`runOrchestratorTask`**: an internal `AbortController` is threaded into the engine and aborted when the timeout wins the race **or** `deps.signal` fires; the caller-abort listener is forwarded and cleaned up in `finally`.
- **`index.ts`**: the `pollOnce` `finally` now clears `currentOrchestratorAbort` alongside the SDK/Ollama slots (leak fix).

## Tests

Red/green TDD — 8 new tests, confirmed failing first:
- `worker-executor`: DirectModel already-aborted short-circuit (no fetch) + in-flight fetch abort; PiHarness pre-aborted no-spawn + mid-run child kill.
- `model-invoker`: `opts.signal` → `WorkerRequest.signal`.
- `engine`: signal forwarded into all 5 invokes.
- `orchestrator-executor`: task timeout aborts the engine signal; `deps.signal` abort forwarded mid-run.

Also updated the pre-existing timeout test to simulate a *real* timeout (signal-driven fetch rejection) now that timeout and cancel are reported distinctly.

Full suite green: **1160 tests pass**, build clean.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -4493,6 +4493,7 @@ async function pollOnce(): Promise<{ hadTask: boolean; queueDepth: number }> {
     stopCancellationWatch();
     currentSdkAbort = null;
     currentOllamaAbort = null;
+    currentOrchestratorAbort = null;
     currentCancellation = null;
     currentTask = null;
     currentTaskConfig = null;
--- a/src/orchestrator/engine.ts
+++ b/src/orchestrator/engine.ts
@@ -162,8 +162,10 @@ export async function runOrchestration(
   taskPrompt: string,
   invoker: ModelInvoker,
   config?: Partial<OrchestratorConfig>,
+  opts?: { signal?: AbortSignal },
 ): Promise<OrchestrationResult> {
   const cfg: OrchestratorConfig = { ...DEFAULT_ORCHESTRATOR_CONFIG, ...config };
+  const signal = opts?.signal;
   const allCosts: (number | null)[] = [];
   const warnings: string[] = [];
   let totalLatencyMs = 0;
@@ -171,7 +173,7 @@ export async function runOrchestration(
   // -------------------------------------------------------------------------
   // 1. Plan
   // -------------------------------------------------------------------------
-  const planResp = await invoker.invoke("planner", buildPlannerPrompt(taskPrompt));
+  const planResp = await invoker.invoke("planner", buildPlannerPrompt(taskPrompt), { signal });
   allCosts.push(planResp.costUsd ?? null);
   totalLatencyMs += planResp.latencyMs;
   if (planResp.ok && planResp.truncated) {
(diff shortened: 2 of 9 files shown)


Magnus-Gille:
## Codex cross-model review

Ran a Codex (`gpt-5.5`, xhigh) review of the branch. No critical/medium. One valid **Low** finding:

- The timeout-vs-cancel reason tracking wasn't first-writer-wins — an edge race between the timeout timer and an external abort could report the wrong reason (`timed out` vs `aborted`), undercutting the exact distinction this PR adds. Codex confirmed the listeners are removed on the completion/error paths, `currentOrchestratorAbort` is cleared in the poll `finally`, and the timeout-losing `enginePromise` doesn't create an unhandled rejection (still an input to `Promise.race`).

Fixed in fbe247e (test-first): both executors now record the reason first-writer-wins via a no-op-once-set helper. The regression test uses fake timers to force the race (external abort at t=0, timeout at t=50, rejection at t=100) and asserts the message stays "aborted". Full suite green — 1161 tests pass.

PR #127, opened 2026-07-02
feat(orchestrator): enable private + sovereign (Berget) lane through the dispatcher (#111)

### LINKED ISSUE #111: orchestrator: enable private + sovereign (Berget) lane — dispatcher rejects private orchestrator tasks before role guard
From the Codex review of PR #108 (finding #2, medium).

The dispatcher-level sensitivity check calls `getDispatcherRuntimeMaxSensitivity(task.runtime)` before entering the orchestrator executor. No registry row has `dispatcherRuntime: "orchestrator"`, so it falls back to `internal` and every `Runtime: orchestrator` + `Sensitivity: private` task is rejected before `assertProvidersAllowSensitivity()` can check whether all roles are bound to a sovereign provider (Berget). The documented all-Berget private lane therefore can't run through the dispatcher.

Note: current behaviour is **safe** (fail-closed = private orchestrator tasks rejected, no data leak) — this is about *enabling* the private lane, not a leak.

**Fix:** special-case `orchestrator` in the dispatcher sensitivity check and defer to `assertProvidersAllowSensitivity()` before any model call (or model `orchestrator` as max-private with the role guard as the enforcing policy). Tests: private + all-Berget roles succeed; private + default OpenRouter makes zero model calls.

Magnus-Gille:
Closes #111.

> **Stacked on #126 → #125** (base = `fix/orch-abort-signal`). GitHub retargets down the stack as each lands. Merge order: #125 → #126 → this.

## Problem

The dispatcher sensitivity gate (`getSecurityViolationForTask`) called `getDispatcherRuntimeMaxSensitivity(task.runtime)`. No `RUNTIME_REGISTRY` row maps `dispatcherRuntime:"orchestrator"`, so the lookup fell back to `"internal"` and **every** `Runtime: orchestrator` + `Sensitivity: private` task was rejected *before* `assertProvidersAllowSensitivity()` could check whether all roles are bound to a sovereign provider (Berget). The documented all-Berget private lane could not run through the dispatcher.

Note: prior behaviour was **safe** (fail-closed — private orchestrator tasks rejected, no leak). This is about *enabling* the private lane, not fixing a leak.

## Fix

Special-case `"orchestrator"` in `getDispatcherRuntimeMaxSensitivity` to return the top of the lattice (`"private"`) and defer enforcement to the role guard, which runs fail-closed at the very start of `runOrchestratorTask` before any model call:
- **private + all-Berget roles** → admitted.
- **private + default OpenRouter** → rejected by `assertProvidersAllowSensitivity`, **zero model calls**.

No leak: the role guard gates every orchestrator model call regardless of prompt/context-ref content, so a private context-ref with a non-sovereign role can never reach a provider.

## Why this seam

`getDispatcherRuntimeMaxSensitivity` is exported and its only caller is `getSecurityViolationForTask` (`index.ts:821`), which treats a falsy return as "allowed". Encoding the policy there keeps it unit-testable (matches how the codebase already tests runtime ceilings in `sensitivity.test.ts`).

## Tests

- Red/green TDD: new `sensitivity.test.ts` case — `getDispatcherRuntimeMaxSensitivity("orchestrator") === "private"`.
- The executor role-guard behavior the issue asks for (private+Berget admitted; private+OpenRouter → `invoke` never called) is already covered in `orchestrator-executor.test.ts`.

Full suite green: **1162 tests pass**, build clean. CLAUDE.md sensitivity-guard note corrected to describe the deferral.


Magnus-Gille:
## Codex cross-model review (security-focused)

Ran a Codex (`gpt-5.5`, xhigh) review with a security emphasis. **No leak path found** — Codex traced the full chain and confirmed a private orchestrator task's data cannot reach OpenRouter despite the loosened dispatcher gate: `runOrchestratorTask` is the only dispatcher entry to the engine; `assertProvidersAllowSensitivity` runs first (before abort setup and before any `invoke`); the change is scoped to `runtime === "orchestrator"`; and the broker `orch-v1` OpenRouter worker is a separate path that only handles `openrouter`-resolved envelopes.

Two non-leak findings, both fixed in 42b57f3 (test-first):

- **Medium** — orchestrator `Context-refs` were gated but never injected (CLAUDE.md promises injection; the Ollama path does it, the orchestrator path passed only `task.prompt`). #111 makes private all-Berget orchestrator tasks with private refs admittable, so silently dropping refs was a real contract gap. Added `OrchestratorTaskInput.injectedContext` (wired from `task.contextResolution`), prepended as a `## Context` section **only after the guard passes** so it never materializes for a rejected task.
- **Low** — added the missing dispatcher→executor handoff regression: `private + OpenRouter + injectedContext` → guard rejects with **zero** model calls; `private + all-Berget + injectedContext` → admitted and context reaches the engine.

Full suite green — 1165 tests pass.

PR #145, opened 2026-07-04
feat(orchestrator): quality-adjusted savings — join savings tracker with verdict outcomes

### LINKED ISSUE #144: Quality-adjusted savings: join savings tracker with verdict outcomes so the economics signal can't reward cheap-but-wrong
## What

Make the orchestrator's economics **quality-adjusted**: join the savings tracker (PR3, #142) with the verdict layer (PR2, #140) so the reported "savings" and any routing decisions derived from them are conditioned on verdict outcome — not raw cost delta.

## Why (vision v0.2 §Open questions → Signal design; §Self-knowing by design)

Signal design is the vision's untouched open question, and as of the 07-04 fleet it has become the binding constraint: loops now close and *act* (adaptive verify in `src/orchestrator/engine.ts`, learned routing table live on the gateway). A wrong signal now optimizes the wrong thing at machine speed.

Concrete failure mode: the savings tracker measures cost-vs-Claude. If cheap-local attempts that later **fail verification or get escalated** still book as "savings" (or their retry/escalation cost isn't attributed back to the local attempt), the metric systematically flatters delegation — and anything that consumes it will learn to prefer cheap-and-subtly-wrong.

## Acceptance

- [ ] Savings records carry the subtask's verdict outcome (pass / fail / unknown / escalated), and escalation + retry costs are attributed to the local attempt that caused them.
- [ ] A headline **quality-adjusted savings** number exists alongside raw savings (e.g. net savings counting failed-then-escalated work at full cost); both visible wherever savings are reported (Heimdall descriptor / status).
- [ ] Whatever consumes savings data for decisions (now or later) reads the quality-adjusted series, not the raw one — documented in the orchestrator README.
- [ ] Red/green test: a scenario where local is cheaper but fails verification must show negative or ~zero net savings.

Refs: hugin #140 (verdict layer), #142 (savings tracker); grimnir `docs/gap-analysis-2026-07-03.md` §1.1; vision v0.2 §Signal design.

Filed by: grimnir

Magnus-Gille:
## What

Joins the savings tracker (PR3, #142) with the verdict layer (PR2, #140) so the orchestrator's economics signal is conditioned on verdict outcome. The raw `savedUsd` series books a cheap-but-wrong local attempt as a win; the new headline `qualityAdjustedSavedUsd` books it as a loss.

## Design decisions (per the issue's open questions)

**Schema extension (write-time join), not a read-time join.** The ordering concern — "the savings store may be written before the verdict is known" — turns out not to apply: verification runs *inside* `runOrchestration` (engine step 3), and the executor computes + records savings only on the completed `OrchestrationResult`, where every verdict is final. So a single write-time join is correct and no two-phase write is needed. The one genuinely verdict-less case is an adaptive-verify skip, which is semantically **"unknown"** for that run (never pending) and is recorded as such — never as a fake pass.

**Attribution back to the local attempt.** `ModelCallRecord` gains `subtaskId` (set for worker and verifier calls; absent for run-level planner/synth). Credit rules in `computeSavings`:

| Call | Outcome | Baseline credit |
|---|---|---|
| worker | `pass` | full |
| worker | `unknown` (unverified) | full, but bucketed separately in `byOutcome` so consumers can discount |
| worker | `fail` / `error` / `escalated` | **zero** — the work still has to be done at the frontier; local spend books as a loss |
| verifier | any | **zero** — the all-Claude counterfactual doesn't verify; verification cost is attributed to the local attempt that caused it, never booked as neutral independent frontier spend |
| planner / synthesizer | run-level | full (raw treatment) |

`escalated` is reserved: the engine has no escalation/retry path today (a failed-verdict subtask is excluded from synthesis, not re-run). The outcome type + credit rule exist now so that when escalation lands, its calls carry the originating `subtaskId` and their cost lands on the causing attempt — documented in `src/orchestrator/README.md` and the ADR.

**Headline number.** `qualityAdjustedSavedUsd = qaBaselineCreditUsd − actualCostUsd` (can be negative). Both series surface everywhere savings already surfaced: the human summary (second line), `result-structured.savings` (additive/optional fields), and the `tasks/_savings` Munin doc (`totals.qaBaselineCreditUsd` + `byOutcome` buckets; lifetime QA savings derive at read time as `totals.qaBaselineCreditUsd − totals.actualCostUsd`, mirroring the store's derive-don't-store convention). Savings do not currently appear in the Heimdall descriptor (confirmed by inspection — the existing ADR lists a Heimdall panel as an explicit non-goal), so there was nothing to extend there; a Heimdall consumer reading `tasks/_savings` gets the QA counters.

**Migration safety.** A pre-#144 `tasks/_savings` doc sanitize-defaults the new fields (`qaBaselineCreditUsd` → 0, `byOutcome` → `{}`) instead of being reset — same precedent as the verdict store's `unverifiedPasses`. `schemaVersion` stays 1.

**Consumption rule (acceptance item c/d).** New `src/orchestrator/README.md`: any consumer using savings data for decisions MUST read the quality-adjusted series, never raw `savedUsd`. Also documented in the ADR (S6) and CLAUDE.md.

## Test evidence

Red/green: the acceptance test (local cheaper but fails verification → ~zero/negative net savings) was written first and failed against the raw-savings logic (`qualityAdjustedSavedUsd` undefined), then went green: raw `savedUsd ≈ +$17.73`, `qualityAdjustedSavedUsd = −$0.27` for the same run. Store and executor-join tests were likewise written red-first (6 red → green).

- `tests/orchestrator/savings.test.ts`: 19 pass (8 new — verifier attribution, unknown/error/escalated, byOutcome buckets, run-level calls, no-map backward compat)
- `tests/orchestrator/savings-store.test.ts`: 21 pass (6 new — QA counters, pre-#144 doc preservation, malformed-row drops, negative-QA-is-valid-data)
- `tests/orchestrator/orchestrator-executor.test.ts`: 48 pass (3 new — end-to-end join through engine outcomes, summary line)
- `tests/orchestrator/engine.test.ts`: subtaskId attribution test
- Full suite: **1420/1420 pass**, `tsc` clean.

Closes #144


Magnus-Gille:
**Codex cross-model review (gpt-5.5, effort high) — 2 findings, both fixed in ef76dd0:**

1. **medium — duplicate planner-emitted subtask ids could collapse verdicts in the savings join.** Confirmed against `parsePlan` (it normalized empty ids but never deduped non-empty ones, and the join is keyed by subtask id). Fixed test-first: ids are now deduped at plan-parse time (first occurrence keeps its id, later duplicates get a deterministic non-colliding suffix), with unit tests plus an end-to-end regression test (duplicate ids, one pass + one fail → `byOutcome` shows both, not two copies of the last verdict).
2. **low — README savings example had irreconcilable numbers.** Reworked into a fully reconciling worked example.

Full suite after fixes: 1423/1423 pass, `tsc` clean.

Magnus-Gille:
## Orchestrator review leg (Fable, second reviewer)

Independently verified the three riskiest claims against source in the worktree:

1. **Ordering** — confirmed: verification completes inside the engine run and savings are computed at write time from finished outcomes with verdicts joined (`orchestrator-executor.ts` renders `byOutcome` + QA headline from the same summary). The feared two-phase-write problem genuinely doesn't arise.
2. **QA math** — confirmed at `savings.ts:133` (verifier calls: zero credit), `:136` (fail/error/escalated: zero credit), `:266` (`qualityAdjustedSavedUsd = qaBaselineCredit − actual`). Matches the issue's 'can't reward cheap-but-wrong' requirement; the red/green acceptance test (−$0.27 QA vs +$17.73 raw) exercises exactly this.
3. **Codex dedup fix** — confirmed at `plan.ts:153-165`: `seenIds` set with deterministic suffixing of planner-emitted duplicates, so verdict joins can't collapse.

**M5 first-pass** (qwen3-coder-next-80b, full diff inlined): NO FINDINGS.

**Flags for the owner (not blockers):**
- The *unknown-keeps-full-credit* choice is the real judgment call — defensible given adaptive-verify only skips evidence-trusted rows, and `byOutcome` bucketing means a later discount needs no schema change. Recommend accepting as-is.
- QA ≤ raw on every verified run (verifier spend never earns credit) — honest, but worth remembering when reading dashboards: enabling verification will *look* like savings got worse.

Verdict: **ready for human merge** (comment only — formal approval stays with Magnus).

PR #168, opened 2026-07-11
Make Broker alias availability truthful

### LINKED ISSUE #166: Broker: advertise and accept only aliases with live executors
## Problem

The MCP Broker advertises and accepts four aliases, but its only worker can execute OpenRouter one-shot tasks. `tiny`, `medium`, and `pi-large-coder` can be durably accepted into an `orch-v1` queue that no worker drains. Even `large-reasoning` is not executable when `OPENROUTER_API_KEY` is absent.

This is a fail-open contract defect and an authenticated queue-exhaustion risk: accepted is presented as runnable when it is not.

## Scope

- [ ] One capability source shared by submission, `/models`, and worker eligibility.
- [ ] Distinguish executor implemented from executor enabled in this process.
- [ ] Reject unavailable aliases before idempotency reservation, Munin writes, or journal events.
- [ ] `/models` advertises only enabled aliases and backing runtime rows.
- [ ] MCP tool schema uses the live set; failed discovery disables submission rather than advertising a compiled fallback.
- [ ] Reject alias-map version skew explicitly.
- [ ] Preserve the broader alias vocabulary for historical journal/envelope reads.
- [ ] Regression tests cover all three unimplemented aliases and disabled OpenRouter.

## Acceptance

An alias visible in `hugin_models` can be claimed by a live worker. Every other alias fails immediately with typed `alias_unavailable` and creates no durable task.

Filed by: Codex from the 2026-07-11 Hugin audit.

Magnus-Gille:
## What changed

- derives one live Broker executor-capability contract from the same OpenRouter configuration that starts the worker
- rejects unimplemented or disabled aliases before idempotency reservation, Munin writes, or journal events
- makes `/v1/delegate/models` return only enabled aliases and their backing runtime rows
- makes hugin-mcp build its submit alias schema from live discovery and disable submission when discovery fails or returns no executor
- rejects alias-map version skew explicitly while retaining all four aliases for historical envelope/journal parsing
- updates the Broker contract and both agent instruction files

## Why

The Broker advertised and accepted four aliases even though only OpenRouter one-shot tasks had a worker. Three aliases could remain pending forever, and `large-reasoning` could do the same when `OPENROUTER_API_KEY` was absent. This made accepted work look runnable and allowed authenticated queue exhaustion.

## Validation

- `npm test` — 91 files, 1518 tests green
- `npm run build`
- `git diff --check`

Closes #166

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -117,6 +117,7 @@ hugin/
 │   └── broker/                   # Orchestrator-v1 broker (Tailscale-only HTTP, /v1/delegate/*)
 │       ├── server.ts             # Express app + opt-in startup (HUGIN_BROKER_KEYS)
 │       ├── handlers.ts           # submit/await/rate/list/models endpoint handlers
+│       ├── executor-capabilities.ts # Live executor truth shared by submit/models/worker
 │       ├── orch-worker.ts        # Polls Munin for orch-v1 tasks, claims via CAS, dispatches to OpenRouter
 │       ├── openrouter-executor.ts # OpenRouter one-shot delegation runner
 │       ├── reconciliation.ts     # Periodic sweep: backfill journal events for orch-v1 tasks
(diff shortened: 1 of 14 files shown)


Magnus-Gille:
Native Codex review completed against 01e8489ecf6eda12416041c0c24c41114cfa25e7 (Claude review intentionally skipped per Magnus).

No blocking findings.

Reviewed for fail-open routing, queue poisoning, startup/config drift, historical wire compatibility, alias-map versioning, MCP discovery failure, idempotency side effects, and documentation drift. The review strengthened one regression test to prove an unavailable-alias rejection writes neither Munin nor journal state and does not consume the idempotency key.

Validation: full suite 91 files / 1518 tests green; focused post-review Broker test green; TypeScript build and diff check clean.

Residual risk: executor availability is configuration/process capability, not an upstream OpenRouter health guarantee; per-call provider failures remain normal typed execution failures.

PR #179, opened 2026-07-12
fix: distinguish missing vs broken claude-config checkout in deploy-pi (#153)

### LINKED ISSUE #153: Make deploy-pi handle missing claude-config checkout cleanly
During the 2026-07-08 deploy for #152, `scripts/deploy-pi.sh` completed successfully but printed:

```
WARNING: claude-config bootstrap failed — clone ~/repos/claude-config on the Pi first, then re-run.
```

The service deploy and health check were fine, but the warning is noisy and ambiguous for routine Hugin deploys. It should either:

- auto-clone/pull the expected `~/repos/claude-config` dependency before running `./bootstrap.sh --no-plugins`, or
- detect that this step is optional for the current deploy and print a clear non-warning informational message with the exact remediation command.

Acceptance:

- A normal `./scripts/deploy-pi.sh` run should not leave an unexplained bootstrap warning when Hugin itself is healthy.
- If claude-config is truly required and missing, the script should explain why and how to fix it.

Filed by: gille-inference / Codex after #152 deployment.

Magnus-Gille:
## Why

`scripts/deploy-pi.sh`'s claude-config bootstrap step printed the same
WARNING regardless of *why* it failed: whether `~/repos/claude-config` was
simply never cloned on the Pi (optional third-party config infra, unrelated
to Hugin's own health — tracked separately as claude-config#2) or an
existing checkout was actually broken (a real problem worth flagging). This
made routine, healthy Hugin deploys look like they had an unresolved issue.

Issue #153 asked for the script to tell apart "optional, not set up" from
"actually required and broken."

## What

Extracted the bootstrap logic into `scripts/lib/claude-config-bootstrap.sh`
(`claude_config_bootstrap()`), sourced by `deploy-pi.sh`:

- **Checkout missing** → prints an `INFO:` line (not a warning) explaining
  the step is optional and was skipped, plus the exact `git clone` command
  to enable it.
- **Checkout present but `git pull`/`bootstrap.sh --no-plugins` fails** →
  still prints a `WARNING:` — something is genuinely broken and needs
  inspecting.
- **Checkout present and healthy** → silent, as before.

Kept the diff surgical: no behavior change to the rest of `deploy-pi.sh`,
no auto-clone (a private repo clone from inside the deploy script isn't
"trivially safe" — it would need credentials/agent forwarding the script
can't assume), just a clearer signal.

## Test plan

- [x] New bash-level test `scripts/lib/claude-config-bootstrap.test.sh`
      stubs `ssh` (no network/remote access) and asserts all three code
      paths (missing / healthy / broken checkout). Written and confirmed
      failing (red) before the implementation existed, then green after.
- [x] `shellcheck` on both changed/new files — no new findings; only the
      same pre-existing info-level `SC2029` notes already present elsewhere
      in `deploy-pi.sh`.
- [x] `bash -n` syntax check on all three files.
- [x] Full test suite: `npm run build` + `npm test` — 95 files / 1612 tests
      pass (this change is bash-only and doesn't touch the TS suite).
- [x] Did **not** execute the actual deploy or ssh anywhere, per scope.

Closes #153

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,3 +25,5 @@ jobs:
       - run: npm run build
 
       - run: npm test
+
+      - run: bash scripts/lib/claude-config-bootstrap.test.sh
--- a/scripts/deploy-pi.sh
+++ b/scripts/deploy-pi.sh
@@ -130,8 +130,11 @@ echo "==> Refreshing Claude config on the Pi (claude-config bootstrap)..."
 # Config now lives in the versioned Magnus-Gille/claude-config repo (+ claude-skills,
 # skills-private), consumed via symlinks. Pull latest + re-run bootstrap instead of the
 # old rsync (which would clobber the symlinks with real dirs). See claude-config/README.md.
-ssh "$REMOTE" 'cd ~/repos/claude-config 2>/dev/null && git pull -q --ff-only && ./bootstrap.sh --no-plugins' \
-  || echo "  WARNING: claude-config bootstrap failed — clone ~/repos/claude-config on the Pi first, then re-run."
+# See scripts/lib/claude-config-bootstrap.sh (issue #153): missing checkout is optional
+# infra and informational, a broken existing checkout stays a real WARNING.
+# shellcheck source=lib/claude-config-bootstrap.sh
+source "$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/lib/claude-config-bootstrap.sh"
+claude_config_bootstrap "$REMOTE"
 
 echo "==> Installing CLI update cron job..."
 CRON_CMD="0 4 * * * $REMOTE_DIR/scripts/update-cli.sh 2>&1 | logger -t hugin-update"
--- a/scripts/lib/claude-config-bootstrap.sh
+++ b/scripts/lib/claude-config-bootstrap.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+# claude-config bootstrap helper for deploy-pi.sh (issue #153).
+#
+# claude-config is optional infra for a Hugin deploy — tracked separately as
+# claude-config#2, not a Hugin concern. A missing checkout is expected and
+# informational, not a deploy WARNING. An existing checkout that fails to
+# pull/bootstrap is a real problem (something broke), so that stays a
+# WARNING with its own distinct message.
+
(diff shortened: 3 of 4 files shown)


Magnus-Gille:
**Codex cross-model review** (gpt-5.6-sol, high reasoning effort — first attempt hit a broken local `codex-code-mode-host` install, resolved with `--disable code_mode_host`).

3 findings, all addressed in 8a65d47:

1. **Medium** — `scripts/lib/claude-config-bootstrap.sh:12-17`: the `else` branch ran for *any* nonzero result from the presence-probe `ssh`, not only a genuine "directory absent." An SSH transport/auth/timeout failure would have been reported as the optional/missing case (INFO), hiding a real connectivity problem. Fixed by distinguishing ssh's own exit-code convention: `1` = the remote `test` ran and correctly reported absence, anything else = ssh itself couldn't run the check and absence must not be assumed (→ WARNING).
2. **Medium** — checking `test -d .../claude-config/.git` rather than the checkout directory itself could misclassify some present-but-broken checkouts. Switched to checking the checkout directory itself; a broken/incomplete checkout now correctly falls through to the pull/bootstrap attempt and its WARNING-on-failure path.
3. **Low** — the new bash test wasn't wired into CI. Added a step to `.github/workflows/ci.yml`. Also added a 4th test case covering the inconclusive-probe path from finding 1.

Verified: red/green TDD on both fixes, `shellcheck` clean (only the same pre-existing info-level notes), full suite still 95 files / 1612 tests green.

Magnus-Gille:

- Verified the ssh presence-probe exit-code logic in `scripts/lib/claude-config-bootstrap.sh`: rc 0/1/other three-way split is the correct ssh convention (remote `test` status vs 255-class transport failure), and the quoting is right — the tilde expands on the remote, not locally. The stubbed-ssh test covers all four outcomes.
- Verified the CI claim: `claude-config-bootstrap.test.sh` is wired into `.github/workflows/ci.yml`.
- Cross-checked the DELEGATIONS report against broker records: all 9 delegated sub-task IDs exist on the broker, all terminal, all rated (9/9). Honest grade distribution (3× wrong, 1× gateway escalation) — good harvest data.
- Open for owner judgment (relayed from the session's uncertainties): the printed remediation command uses the SSH git URL (`git@github.com:…`) — confirm that matches how the Pi authenticates for private clones.

Issue #129, opened 2026-07-03
Overnight autonomous tasks silently killed by expired Pi Claude auth (401) — no pre-flight check, cause buried in raw log


Magnus-Gille:
## Summary

An overnight autonomous Hugin research-spike task failed at startup with a **401 authentication error** on the Pi's Claude Code agent-SDK runtime, producing nothing. The task ran for 9 seconds ($0.00, 1 turn) and was marked `failed`. Because auth is checked only at the first API call — *after* MCP servers connect and the task is already "started" — the failure is silent from the outside: the task simply disappears, and any downstream work that assumed its output (a morning report, in this case) is blocked with no visible cause.

## Evidence

Task: `20260702-181856-memory-evolvability` (research spike, runtime `claude`, model `claude-opus-4-6`, timeout 5400000ms).

From `~/.hugin/logs/20260702-181856-memory-evolvability.log` on huginmunin:

```
Started: 2026-07-02T18:21:29.257Z
[system] init ... "apiKeySource":"none" ... mcp_servers all "connected"
Failed to authenticate. API Error: 401 {"type":"error","error":{"type":"authentication_error","message":"Invalid authentication credentials"},"request_id":"req_011CcdhH87LLqmvP9Fy5gbUe"}
[SDK Error: Claude Code returned an error result: Failed to authenticate. API Error: 401 ...]
Exit code: 1
Duration: 9s
Cost: $0.0000
Turns: 1
Completed: 2026-07-02T18:21:38.351Z
```

Note `"apiKeySource":"none"` in the init event — the runtime launched with no valid credential.

## Impact

- The overnight research spike produced neither of its two contracted artefacts. Downstream ("MORNING: read overnight spike report") had nothing to read.
- This is an **environment/credential** failure, not a task-logic bug — but the way Hugin surfaces it makes it look like a generic task failure, so the real cause (expired Pi Claude auth) isn't obvious without reading the raw log.
- Any overnight autonomous task will hit the same wall until the Pi's Claude credential is refreshed — a single expired token silently drains the whole overnight queue.

## Asks (for hugin's owner to triage)

1. **Restore the Pi's Claude Code credential** so autonomous tasks can authenticate (immediate unblock).
2. **Pre-flight auth check before dispatch** — validate the runtime can authenticate (cheap ping) *before* marking a task `started`/consuming its slot, so an auth failure is caught early and reported distinctly from a task-logic failure.
[...]

PR #180, opened 2026-07-12
feat(worker): version-drift self-check for on-disk SDK/binary mismatch

### LINKED ISSUE #123: worker: version-drift self-check — detect loaded-vs-on-disk SDK/binary mismatch
## Context

Mirrors the defense-in-depth decision on **grimnir#31**. The 2026-06-17 incident: a `deps-bump-all` task upgraded `@anthropic-ai/claude-agent-sdk` (and its native arm64 `claude` binary) in `~/repos/hugin` **while the Hugin worker kept running**. The worker's in-memory (old) SDK then spawned the new, incompatible binary → every agent-sdk task died instantly with an opaque `exit 1`, silently breaking all autonomous tasks for ~2 days. Fresh processes always worked (they load the current on-disk SDK), which masked the cause as an auth/env problem.

`deploy.sh` already restarts after `npm install`, so the deploy path is safe — the gap is **out-of-band** dependency bumps under a live worker. The cross-cutting fix chosen in grimnir#31 is a per-service self-check; Hugin is the priority since it was the victim.

## Scope

At worker startup (and optionally a periodic re-check), compare the **loaded** version/identity of critical deps against what's **on disk**:
- `@anthropic-ai/claude-agent-sdk` package version
- its native `claude` binary (path + version)

On mismatch: emit a Telegram alert (via the shared notify path) and either refuse to spawn agent-sdk tasks or hard-fail fast with a clear, non-opaque error pointing at "restart the worker — deps changed under it."

## References
- grimnir#31 (decision + rationale)
- hugin#114 (sdk-executor now captures child stderr — makes this class diagnosable)

Magnus-Gille:
## Summary

- Fixes issue #123: the worker now snapshots the on-disk `@anthropic-ai/claude-agent-sdk` version and its vendored `cli.js` runtime identity (path/size/mtime) once at startup, and compares against a fresh on-disk read before every `claude`-runtime agent-sdk task (`src/version-drift.ts`, pure build/compare logic).
- On a confirmed mismatch — deps bumped on disk while the worker kept running, the 2026-06-17 incident class — the task is refused fast with a distinctly-classified `DEPS_DRIFT` failure (`failure:deps-drift` tag, `- **Failure kind:** DEPS_DRIFT`) instead of an opaque `exit 1`, and one edge-triggered critical alert is pushed through the existing Ratatoskr Alert Bus path (reusing `sendRatatoskrAlert`, the same send-target env vars as `HUGIN_AUTH_ALARM`).
- The check itself is fail-open by design: any error reading on-disk state (missing baseline, mid-`npm install` partial write, permissions) is logged and treated as no-drift — only a genuine, confirmed mismatch ever blocks a task.
- New env var `HUGIN_VERSION_DRIFT_CHECK` (default `on`). `failure-classification.ts` gains `DEPS_DRIFT_FAILURE_KIND`/`TAG` so the synthetic short-circuit output flows through the same Munin tagging/structured-result pipeline as the existing `AUTH_FAILED` classification.

## Why

- The 2026-06-17 incident: a `deps-bump-all` task upgraded the SDK/binary on disk while the worker kept running, silently breaking every agent-sdk task for ~2 days behind an opaque `exit 1`.
- `deploy.sh` already restarts after `npm install`, so the deploy path is safe — the gap is out-of-band dependency bumps under a live worker.
- Mirrors the defense-in-depth decision on grimnir#31; Hugin is the priority service since it was the victim.

## Test plan

- [x] Full test suite passes: `npx vitest run` — 1629/1629 passed (96 test files).
- [x] `npx tsc --noEmit` clean.
- [x] `npm run build` clean; verified `resolveClaudeSdkDir()`/`readOnDiskVersionSnapshot()` resolve correctly from the built `dist/index.js` (confirmed the SDK package's `exports` map blocks direct `./package.json`/`./cli.js` subpath resolution — fixed by resolving the bare package specifier and taking `path.dirname()`).
- [x] New unit tests for `src/version-drift.ts` (14 tests, TDD'd red→green) and 3 new `failure-classification.ts` tests for the `DEPS_DRIFT` classification path.
- [x] Manually smoke-tested `resolveClaudeSdkDir`/`readOnDiskVersionSnapshot` end-to-end against the real installed SDK package from the built `dist/` output.
- [ ] Not verified: an actual live Ratatoskr alert delivery, or a real on-disk mutation triggering the drift path end-to-end on a running worker — both require infra outside this dev sandbox.

Closes #123


Magnus-Gille:
Ran a Codex cross-model review (`gpt-5.6-sol`, `model_reasoning_effort=high`) on this branch. 4 findings, all addressed:

**1. [Medium, fixed] DEPS_DRIFT misreported to the auth alarm.** `classifyClaudeFailure()` now returns either `AUTH_FAILED` or `DEPS_DRIFT`, but the finalize path treated *any* non-null classification as auth-failure evidence — a confirmed drift refusal would log "Task failed to authenticate (DEPS_DRIFT)" and call `noteClaudeAuthOutcome("unauthorized")`, wrongly flipping the Pi-credential auth alarm. Fixed in 009434c: the auth-alarm call is now gated on `failureClassification?.kind === AUTH_FAILURE_KIND` specifically, and the log message no longer hardcodes "authenticate".

**2. [Medium, fixed] DEPS_DRIFT was regex-inferred from output text.** The only signature was `/\bDEPS_DRIFT:/` applied to *any* failed Claude task's captured output — a legitimate task that happened to print that literal token for an unrelated reason (e.g. one working on this very feature) could get mis-tagged `failure:deps-drift`. Fixed in 009434c: DEPS_DRIFT is no longer inferred from output at all. The pre-flight short-circuit already knows with certainty it refused the task, so it now returns a trusted `preflightFailureKind` discriminator that the finalize path consults directly via the new `driftFailureClassification()`. `classifyClaudeFailure()` only ever infers `AUTH_FAILED` now (unchanged behavior, still needed for real runtime 401s that aren't pre-flight-caught). Added a regression test proving the literal-token case classifies as `null`, not `DEPS_DRIFT`.

**3. [Low, deferred] No direct unit tests for the fs/preflight IO glue in index.ts.** Correct gap, but this matches the existing codebase convention: `index.ts`'s deep dispatcher internals (e.g. `probeClaudeUsage`, the pre-existing `#129` auth-preflight short-circuit) are exercised indirectly, not unit-tested directly — only the extracted pure modules (`auth-alarm.ts`, `failure-classification.ts`, now `version-drift.ts`) get direct coverage. A full DI refactor of the IO glue to make it independently testable is a reasonable follow-up but is a larger, separate structural change I'm not taking on in this ticket. The concrete bug this gap let through (finding #1) is now fixed and, more importantly, structurally prevented going forward: `driftFailureClassification()` is a trusted, non-regex path, so the failure mode that caused finding #1 (conflating two failure kinds inferred through the same regex funnel) can't recur the same way.

**4. [Low, pushing back] `AGENTS.md` missing `HUGIN_VERSION_DRIFT_CHECK`.** `AGENTS.md` is a separately-maintained file, not generated from `CLAUDE.md`, and it's already substantially stale independent of this PR — e.g. it's missing the entire `orchestrator`/`pipeline`/`auto` runtimes, the `orin` ollama host, `Sensitivity`/`Capabilities` task fields, and even `HUGIN_CLAUDE_AUTH_PREFLIGHT` from the *existing* #129 work it's supposedly documenting. Reconciling `AGENTS.md` with `CLAUDE.md` wholesale is a pre-existing, repo-wide doc-drift problem well outside this ticket's scope — not something introduced or meaningfully worsened by this change. Not fixing it here; flagging for a separate cleanup pass.

Full suite green after the fixes: 1629/1629, `tsc --noEmit` clean.

Magnus-Gille:

- Independently verified both Codex-medium fixes: (1) `DEPS_DRIFT` is applied only by the pre-flight short-circuit that itself refused the task — `classifyClaudeFailure` remains AUTH-only regex, so arbitrary task output cannot spoof or overwrite a real failure classification; (2) drift alerting is its own edge-triggered channel (`hugin-version-drift` dedup key), fully decoupled from the Pi-auth alarm state.
- Fail-open contract verified at both read sites (startup baseline + per-task re-read): a check error can never block a task; only a confirmed mismatch refuses agent-sdk work.
- Cross-checked DELEGATIONS against broker records: all 9 sub-task IDs exist and are terminal, 9/9 rated. Note for the harvest ledger: `mcp-m5-df675c5b…` is broker-outcome `failed` and was honestly rated `partial(discarded)`.
- Agree with the session's pushback on the AGENTS.md finding (pre-existing, repo-wide) and with deferring the index.ts IO-glue DI refactor — both reasonable headless-session judgment.
- Known accepted tradeoff (session flagged it): `cliSizeBytes`+`cliMtimeMs` instead of a content hash — a same-size-same-mtime swap evades it; documented and acceptable for this threat model.

Issue #165, opened 2026-07-11
Run Hugin role-validation trial and decide keep/reduce/remove on 2026-08-22


Magnus-Gille:
## Goal

Run the canonical Hugin role-validation trial through **2026-08-22** and make an explicit keep / reduce / remove decision from observed use rather than architectural preference.

## Canonical role under test

Hugin is an optional durable macro-broker and supervisor between L1 conductors and execution nodes. It is not the default planner, a required hop for direct M5 access, or a second capability ledger.

## Trial evidence

- [ ] At least 10 substantive Hugin tasks.
- [ ] At least 2 independent producers or recurring workflows.
- [ ] At least 70% useful completion (`pass` or genuinely useful `partial`).
- [ ] At least 5 tasks complete after the initiating L1 session closes.
- [ ] Zero lost tasks, duplicate side effects, or security-boundary failures.
- [ ] Less than 2 hours of maintenance during the trial.
- [ ] Track human rescue/redo, maintenance time, and incidents—not just terminal completion.

## Decision rules

- **Keep core:** the full gate above passes.
- **Reduce:** 3–9 substantive tasks or 40–70% useful completion, with at least one recurring producer.
- **Remove from the active path:** fewer than 3 substantive tasks, no recurring producer, unacceptable reliability/security, or maintenance cost above the gate.
- The mini-Conductor must additionally earn at least 5 substantive uses and a demonstrated cost, availability, or repeatability advantage; otherwise remove it even if the durable supervisor core stays.

## Freeze during trial

Do not expand planner/fanout/synthesis, skill RouteBindings, Hugin capability verdicts, or Hugin routing intelligence unless required to collect valid trial evidence or fix a safety/reliability defect.

## Exit criterion

A dated decision record on 2026-08-22 names what stays, what is reduced, what is deleted, and the evidence behind that decision.

Filed by: Codex from the 2026-07-11 architecture reconciliation.

Magnus-Gille:
Trial baseline note from the #168 deployment: the production Pi currently runs the durable legacy dispatcher but has no `HUGIN_BROKER_KEYS(_FILE)` or `OPENROUTER_API_KEY`, so the MCP Broker endpoint is disabled. Broker/MCP tasks therefore cannot count toward the #165 trial until an intentional producer and credentialed intake are provisioned. Do not infer product adoption from code being deployed.

Magnus-Gille:
Trial progress audit on 2026-07-30:

[...]

PR #182, opened 2026-07-12
fix(broker): stop hugin_list under-counting tasks crowded out by tag pollution

### LINKED ISSUE #181: hugin_list under-counts real broker tasks — recent tasks invisible even with explicit since_ts
Found 2026-07-12 by the m5h-2026-07 fleet's delegation cross-check (orchestrator integrity leg).

## Symptom

The gi#200 fleet session reported 9 delegations. `hugin_list` showed only 5 of them; `hugin_list --since_ts 2026-07-12T13:03:00Z` returned **total: 1** (`mcp-m5-89c8ddfde15a56bb92b921d7`, 13:03:43Z) even though 4 more tasks were submitted after that time. Ground truth via Pi execution logs (`~/.hugin/logs/`): **all 9 exist and executed** — `mcp-m5-d20b22ced4a5b0817a2d3549`, `mcp-m5-2a4b5fd0d69c95402496f6df`, `mcp-m5-051b2740e905c1e2e4c80c33`, `mcp-m5-64a6d68aca1bf85101346c91` are the four invisible ones.

Additional data point: an earlier `hugin_list limit:10` (~13:25Z) returned `total: 9` and had silently dropped several 12:22–12:23Z tasks that a `limit:30` call at ~12:40Z **did** return (e.g. `mcp-m5-342dfd2831c07657e9f9eb13`), while keeping adjacent ones from the same minute (`mcp-m5-d592641c…` 12:23:50, `mcp-m5-df675c5b…` 12:23:09). So visibility is not a simple recency window — real terminal tasks fall out of the listing non-monotonically over time.

## Why it matters

- Nearly caused a false fabrication accusation against an honest session during evidence cross-checking.
- Anything counting broker tasks through the same query path risks under-counting the #165 trial corpus. (CLAUDE.md already warns the `broker:mcp-v2` tag was lossy pre-#173 and the health panel keys on embedded envelopes — `hugin_list`'s filter may have a sibling of that bug, possibly interacting with rating-event writes or terminal-status tag rewrites.)

## Acceptance

- `hugin_list` with `since_ts` returns every canonical broker task for the principal in that window, regardless of rating events or terminal-status normalization.
- A red/green test reproducing the drop (task visible → rate/normalize → task still visible).
- Verify `learning-loop-collector.ts` does not share the defect (it should count by embedded envelope; confirm with a test).

Magnus-Gille:
## Summary
- Fixed `hugin_list` under-counting real broker tasks by ensuring status entries are not displaced by feedback or await-observation entries in Munin's capped result window.
- Refactored `BrokerTaskStore.listCanonical` to union two tag queries (`broker:mcp-v2` and `runtime:homeserver`) and deduplicate by namespace, then fetch status entries directly — eliminating reliance on raw query window survival.
- Threads `since_ts` through to Munin's native `since` filter for accurate and efficient time-based filtering, and reads namespace status entries concurrently since the union roughly doubles the read fan-out.
- `listCanonical` always queries Munin at its real per-query cap (50) regardless of the caller's desired output row count — a small `?limit=` no longer narrows the raw candidate window (Codex review finding, see below).
- Added tests covering both crowd-out scenarios, `since_ts` forwarding, the always-max-query-window invariant, a tag-divergent test proving the `runtime:homeserver` union is load-bearing, the full submit→rate→still-visible acceptance scenario, and a regression test confirming `learning-loop-collector.ts` does not share the defect.

## Root cause
`BrokerTaskStore.listCanonical` queried Munin for entries tagged `broker:mcp-v2`, which is shared by a task's `status`, `feedback` (`hugin_rate`), and `await-observation` (poll) entries. Since Munin caps query results at 50 server-side, once a task was rated or polled, its own or another task's sibling entries could crowd its `status` entry out of the returned window — even though the status entry itself never changed. The old code discarded any raw result whose key wasn't `status`, so a crowded-out task vanished from `hugin_list` entirely, including under an explicit `since_ts` filter, since filtering happened *after* the already-truncated fetch.

## Test plan
- [x] Reproduced crowd-out by a co-tagged `feedback` entry in `tests/broker/task-store.test.ts`
- [x] Reproduced crowd-out by a co-tagged `await-observation` entry in `tests/broker/task-store.test.ts`
- [x] Confirmed `since_ts` is forwarded to both underlying Munin queries in `tests/broker/task-store.test.ts`
- [x] Confirmed the raw query window is always Munin's real max, independent of the caller's `limit`, in `tests/broker/task-store.test.ts`
- [x] Confirmed the `runtime:homeserver` union (not just dedup) is what surfaces a task fully crowded out of the `broker:mcp-v2` channel, in `tests/broker/task-store.test.ts`
- [x] Reproduced the full acceptance scenario (submit → rate → task stays visible under `since_ts`) in `tests/broker/handlers.test.ts`
- [x] Confirmed `learning-loop-collector.ts` already counts by embedded envelope and does not share the defect, in `tests/learning-loop-collector.test.ts`
- [x] All new tests confirmed red against the old/unfixed code before each fix, green after
- [x] Full test suite passes (1636 tests across 96 files)
- [x] Build (`tsc`) clean

## Cross-model review
Reviewed by Codex (`gpt-5.6-sol`, high reasoning effort). Two real findings, both fixed in a follow-up commit (see above): (1) small-`limit` window narrowing, (2) crowd-out tests not actually proving the union was load-bearing. One residual finding — Munin's query API has no cursor/offset pagination, so total enumeration still tops out around ~100 distinct namespaces for a corpus larger than that — is out of scope for this bug-fix PR (current production corpus is ~9-30 tasks/day) and filed as #183.

Closes #181

### DIFF
--- a/src/broker/handlers.ts
+++ b/src/broker/handlers.ts
@@ -482,7 +482,11 @@ export function createListHandler(deps: BrokerHandlerDependencies) {
       res.status(500).json({ error: "internal", message: "principal missing" });
       return;
     }
-    const canonical = await deps.taskStore.listCanonical(principal, parsed.limit ?? 50);
+    const canonical = await deps.taskStore.listCanonical(
+      principal,
+      parsed.limit ?? 50,
+      parsed.since_ts,
+    );
     const historical = Array.from(projectDelegations(await deps.journal.readAll()).values())
       .filter((row) => row.envelope?.broker_principal === principal);
     const canonicalIds = new Set(canonical.map((row) => row.task_id));
--- a/src/broker/handlers.ts
+++ b/src/broker/handlers.ts
@@ -482,11 +482,7 @@ export function createListHandler(deps: BrokerHandlerDependencies) {
       res.status(500).json({ error: "internal", message: "principal missing" });
       return;
     }
-    const canonical = await deps.taskStore.listCanonical(
-      principal,
-      parsed.limit ?? 50,
-      parsed.since_ts,
-    );
+    const canonical = await deps.taskStore.listCanonical(principal, parsed.since_ts);
     const historical = Array.from(projectDelegations(await deps.journal.readAll()).values())
       .filter((row) => row.envelope?.broker_principal === principal);
     const canonicalIds = new Set(canonical.map((row) => row.task_id));
--- a/src/broker/task-store.ts
+++ b/src/broker/task-store.ts
@@ -210,23 +210,52 @@ export class BrokerTaskStore {
     );
   }
 
-  async listCanonical(principal: string, limit = 50): Promise<Array<{
+  /**
+   * Enumerate canonical broker tasks for a principal.
+   *
+   * Deliberately does NOT trust the raw results of a single `broker:mcp-v2`
+   * tag query. That tag is shared by a task's `status` entry AND its
+   * `feedback` (hugin_rate) and `await-observation` (poll) entries, and the
+   * query is capped server-side — so once a task is rated or polled, its own
+   * or another task's sibling entries can crowd the `status` entry out of the
+   * returned window even though nothing about the status entry changed
+   * (#181). Mirrors learning-loop-collector.ts's approach: use whatever
+   * matched (any key, either tag) only to learn the task's *namespace*, union
+   * with a `runtime:homeserver` tag query (carried only by `status` entries,
+   * so it is immune to feedback/await-observation pollution), then read each
+   * unique namespace's `status` entry directly rather than relying on it
+   * having survived inside the raw query window.
+   */
+  async listCanonical(principal: string, limit = 50, sinceTs?: string): Promise<Array<{
     task_id: string;
     submitted_at: string;
     outcome: "completed" | "failed" | "cancelled" | "running";
     alias: DelegationEnvelope["alias_requested"];
   }>> {
-    const { results } = await this.munin.query({
+    const queryLimit = Math.min(50, Math.max(limit * 2, limit));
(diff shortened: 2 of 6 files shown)


Magnus-Gille:

- Root cause verified in the diff and it explains the field evidence exactly: co-tagged sibling entries (notably `hugin_rate` feedback writes — the harvest campaign's own rating traffic) crowd a task's `status` entry out of the raw Munin query window, which was silently capped at 50 server-side while the code believed it could request up to 500. The `runtime:homeserver` tag-union recovery is the right shape since only status entries carry that tag.
- Codex's medium was real and the fix is correct: the raw candidate window is now pinned to `MUNIN_QUERY_MAX` regardless of caller `limit`, with output size enforced downstream — `?limit=1` can no longer reintroduce the crowd-out.
- Regression coverage includes the tag-divergent union test and the learning-loop-collector check (the #165 trial-corpus counting path), per the issue's acceptance criteria.
- Delegation cross-check: **9/9 confirmed real and rated** against Pi execution logs.
- Operational note for the owner: the running Pi broker still under-counts until this is deployed — merge should be followed by a `deploy-pi` run.

Magnus-Gille:
**Documented Sol review** (gpt-5.6-sol, effort=high, orchestrator-run — the session's Codex pass didn't record its model, so this makes the cross-model gate explicit before merge): 2 findings, both residual facets of the pre-existing Munin pagination ceiling rather than defects introduced here — details posted to #183, which now carries them as acceptance criteria. Merging as a strict improvement over the live under-counting behavior.

Issue #164, opened 2026-07-11
Add a learning-loop health panel for capability and product evidence


Magnus-Gille:
## Goal

Expose whether the learning loop is genuinely online and whether Hugin itself is earning its place.

## Two evidence planes

1. **M5 capability evidence:** verified samples by task type × model × node, evidence maturity, promotions/demotions, regressions, unverified backlog, policy/judge version.
2. **Hugin product evidence:** substantive task count, producer/workflow count, useful completion, L1-closed completion, human rescue/redo, maintenance time, incidents, delivery reliability.

Do not collapse these into one generic verdict. M5 answers “can this execution cell do the task?”; Hugin product evidence answers “was durable delegation useful enough to keep?”

## Scope

- [ ] Add a content-blind learning-loop health surface and Heimdall panel.
- [ ] Show denominators and sample maturity; no percentage without `n`.
- [ ] Separate verified, shadow, unverified, and aspirational states.
- [ ] Show the current route-policy and judge-policy versions.
- [ ] Show at least one explainable route change caused by verified evidence, or explicitly state that routing is still shadow/manual.
- [ ] Support the 2026-08-22 Hugin role-validation decision.

## Acceptance

The panel makes it obvious whether evidence is arriving, whether it is strong enough to act on, whether a route changed because of it, and whether Hugin delegations were actually useful.

Filed by: Codex from the 2026-07-11 architecture reconciliation.

PR #185, opened 2026-07-12
feat(broker): soft-require a verifier or rubric for judgment task types (#184)

### LINKED ISSUE #184: hugin_submit: soft-require a verifier or rubric for judgment-flavored task types
Evidence from the m5h-2026-07 harvest: judgment-flavored delegations graded without an explicit rubric produce grader-noise "wrong"s (e.g. `mcp-m5-6411c99c…` — classify graded wrong on a defensible answer; the session itself flagged the subjectivity), and 45% of non-escalated leaves landed as `unverified` capability evidence because acceptance defaulted to `l1_review`.

## Ask
When `task_type` ∈ {classify, qa-factual, triage, memory-decision, claim-verify} AND acceptance is the default `l1_review` AND the prompt contains no rubric section, the `hugin_submit` response should include a **non-blocking warning** field ("judgment-type task submitted without verifier or rubric — capability evidence will be weak"). Document the rule in the MCP tool description so delegators see it at tool-selection time. Fail-open: never reject the task.

Optional: count rubric-less judgment submissions in the learning-loop health panel (#164) so the discipline is observable.

Magnus-Gille:
## Summary

`hugin_submit` now returns a non-blocking `warnings` array in the submit response when all three conditions hold: `task_type` is judgment-flavored (`classify`, `qa-factual`, `triage`, `memory-decision`, `claim-verify`), `acceptance` is the default `l1_review`, and the prompt lacks a rubric section. The warning text is: `"judgment-type task submitted without verifier or rubric — capability evidence will be weak"`. This is purely advisory and never rejects tasks, addressing grader-noise issues in internal harvest projects (see #184).

*(Summary paragraph above was drafted by an M5-delegated `rewrite` task against my longer draft, reviewed and used as-is — see the delegation campaign note at the bottom.)*

## Implementation

- **`src/broker/submit-warnings.ts`** (new) — pure `computeSubmitWarnings()`.
  Rubric detection is deliberately dumb per the issue's ask: a
  case-insensitive substring match for `rubric` or `grading criteria`,
  no NLP.
- **`src/broker/types.ts`** — additive optional `warnings: string[]` field
  on `submitResponseSchema`. No existing field changed.
- **`src/broker/handlers.ts`** — `createSubmitHandler` computes the warning
  once per request (same payload backs all three success paths: fresh 202,
  and both idempotent-reuse 200s) and attaches it only when non-empty.
- **`src/mcp/tools.ts`** — `hugin_submit`'s tool description documents the
  rule so a delegating agent sees it at tool-selection time. No code change
  was needed for pass-through: the MCP submit handler already relays the
  broker's JSON response verbatim (`withIdempotencyKey` spreads the existing
  object), so `warnings` flows to the calling agent unmodified — verified
  with a dedicated test.

Deliberately out of scope (issue marks it "Optional"): counting rubric-less
judgment submissions in the learning-loop health panel (#164). Left for a
follow-up if the discipline needs to be observable on the dashboard.

## Test plan

Red/green TDD — all new tests were written first and confirmed failing
before implementation:

- [x] `tests/broker/submit-warnings.test.ts` (new, 11 tests) — pure-function
      coverage: warns for each of the 5 judgment types; no warning with an
      explicit verifier; no warning with a `## Rubric` heading; no warning
      with a "grading criteria" mention; no warning for a non-judgment
      `task_type`; case-insensitivity of the dumb matcher.
- [x] `tests/broker/handlers.test.ts` (+6 tests) — full HTTP-level
      `POST /v1/delegate/submit` coverage for the same four scenarios
      through the real Express app + Zod schema, plus two tests (added
      after self-review, see below) proving `warnings` also propagates on
      both idempotent-reuse 200 paths, not just the fresh 202.
- [x] `tests/mcp/tools.test.ts` (+1 test) — confirms `hugin_submit`'s
      handler passes a broker-returned `warnings` array through unchanged.
- [x] Full suite: `npx vitest run` → **1654/1654 passing** (97 files).
- [x] `npx tsc --noEmit` and `npm run build` clean.

## Review

Codex CLI was unavailable locally (missing `code-mode-host` binary — a
broken install, not a credits issue), so this PR went through an
adversarial self-review instead (see PR comment below for the full
writeup). One confirmed medium finding — missing coverage on the two
idempotent-reuse 200 paths — was fixed test-first; two low/informational
findings were declined with rationale. Flagging for a real Codex pass once
the local install is fixed.

## Delegation campaign note

Several sub-tasks on this ticket (classify the issue, extract its
acceptance criteria, a factual Q&A against the actual Zod schema, a
data-transform over the task-type taxonomy, a regex derivation, a summary,
this PR-body rewrite, a unit-test-gen pass, an independent `code-implement`
of `computeSubmitWarnings`/`promptHasRubric` from a blind spec, and a
`code-review` of the full diff) were routed through the M5 local-inference
box via `hugin_submit`/`hugin_await`/`hugin_rate` as part of an ongoing
harvest of local-model capability evidence, each independently verified and
rated. The blind-spec `code-implement` output was verified to pass all 11
of this PR's own unit tests; it wasn't merged in favor of the version
already written here, but it's strong capability evidence.

Closes #184

### DIFF
--- a/src/broker/handlers.ts
+++ b/src/broker/handlers.ts
@@ -40,6 +40,7 @@ import {
 } from "./types.js";
 import type { AwaitLifecycle } from "./await-observation.js";
 import type { AuthenticatedRequest } from "./auth.js";
+import { computeSubmitWarnings } from "./submit-warnings.js";
 
 export interface BrokerHandlerDependencies {
   taskStore: BrokerTaskStore;
@@ -109,6 +110,9 @@ export function createSubmitHandler(deps: BrokerHandlerDependencies) {
       respondZodError(res, err);
       return;
     }
+    // #184: same request payload on every success path below (fresh accept
+    // or any idempotent reuse), so computed once and reused verbatim.
+    const submitWarnings = computeSubmitWarnings(request);
 
     if (request.envelope_version !== 2) {
       res.status(400).json({
@@ -217,6 +221,7 @@ export function createSubmitHandler(deps: BrokerHandlerDependencies) {
         task_id: taskId,
         received_at: persistedEnvelope.received_at,
         reused_idempotency: true,
+        ...(submitWarnings.length > 0 ? { warnings: submitWarnings } : {}),
       });
       return;
     }
@@ -228,6 +233,7 @@ export function createSubmitHandler(deps: BrokerHandlerDependencies) {
         task_id: idemOutcome.task_id,
         received_at: nowFn(deps)().toISOString(),
         reused_idempotency: true,
+        ...(submitWarnings.length > 0 ? { warnings: submitWarnings } : {}),
       });
       return;
     }
@@ -274,6 +280,7 @@ export function createSubmitHandler(deps: BrokerHandlerDependencies) {
       task_id: taskId,
       received_at: envelope.received_at,
       reused_idempotency: false,
+      ...(submitWarnings.length > 0 ? { warnings: submitWarnings } : {}),
     });
   };
 }
--- a/src/broker/submit-warnings.ts
+++ b/src/broker/submit-warnings.ts
@@ -1,0 +1,46 @@
+/**
+ * Non-blocking submit-time warnings (issue #184).
+ *
+ * Judgment-flavored task types (classify, qa-factual, triage,
+ * memory-decision, claim-verify) have no mechanical acceptance check to
+ * fall back on — with the default `l1_review` acceptance and no rubric in
+ * the prompt, the resulting pass/fail grade is pure L1 subjectivity, and
+ * the m5h harvest evidence it produces is weak. This never rejects a task;
+ * it only surfaces advice in the submit response.
+ */
+
+import type { DelegationRequest, TaskType } from "./types.js";
+
+export const JUDGMENT_TASK_TYPES: readonly TaskType[] = [
+  "classify",
+  "qa-factual",
+  "triage",
+  "memory-decision",
+  "claim-verify",
+];
+
+export const NO_RUBRIC_WARNING =
+  "judgment-type task submitted without verifier or rubric — capability evidence will be weak";
+
+/**
+ * Deliberately dumb, no NLP: a case-insensitive substring match for
+ * "rubric" or "grading criteria" anywhere in the prompt. False negatives
+ * (a rubric phrased some other way) are expected and fine — this only
+ * gates an advisory warning, never the task itself.
+ */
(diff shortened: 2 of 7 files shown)


Magnus-Gille:
## Cross-model review status

**Codex CLI was unavailable** in this environment: `codex login status` succeeded, but every invocation requiring actual tool use (Read/shell/apply_patch) failed with `failed to spawn code-mode host /opt/homebrew/bin/codex-code-mode-host: No such file or directory` — confirmed the binary is genuinely missing (`test -f` fails). This is a broken local install, not a credits/quota issue.

Per the skill's fallback protocol, ran an **adversarial self-review** instead (explicitly labeled as a substitute, not the real cross-model check — flagging this PR for a real Codex pass once the local install is fixed).

### Findings and resolution

1. **Medium — test coverage gap** (confirmed, fixed in 3a2fcb7): the four `#184` warnings tests only exercised the fresh-202 submit path. The two idempotent-reuse 200 paths (in-memory `IdempotencyIndex` retry, and persisted-envelope reuse after a Broker restart) had no judgment-type coverage proving `warnings` propagates there too. Added two tests covering both — both passed immediately against the existing implementation (confirms the design assumption that all three response paths read from the same freshly-parsed `request`, which the idempotency-hash guarantee ties to an identical `task_type`/`acceptance`/`prompt`), so this was a coverage gap, not a latent bug.

2. **Low — declined**: `computeSubmitWarnings(request)` in `handlers.ts` is computed right after schema parsing, before several checks that can still 400/403/409/503 out, discarding the computed value on those paths. Cost is one `.includes` over 5 items + one regex test — negligible, and moving it later would just duplicate the call at three response sites instead of one. Not worth the complexity.

3. **Low/informational — declined (out of scope)**: `submitResponseSchema` isn't parsed/validated anywhere at runtime — `handlers.ts` builds response objects as ad hoc literals. Pre-existing gap, not introduced by this PR; a schema/handler drift issue if it exists predates #184 and is out of scope here.

Magnus-Gille:
**Documented Sol review** (gpt-5.6-sol, effort=high, orchestrator-run — the session's local Codex install is broken, missing `code-mode-host`): **NO FINDINGS** on correctness, fail-open, wire-contract additivity, or coverage. Combined with the session's adversarial self-review (1 medium fixed: idempotent-reuse coverage) and my independent checks (10/10 delegations verified on Pi logs; warnings field additive on all three response paths), this PR meets the merge bar.

PR #188, opened 2026-07-12
fix(deploy-pi): hard-fail on symlinked node_modules, harden rsync exclude

### LINKED ISSUE #187: deploy-pi: node_modules symlink bypasses rsync exclude and partially deletes remote dependencies
## Incident

During the 2026-07-12 deployment of merged PR #186 from a clean worktree, local `node_modules` was a symlink to the owning checkout. `scripts/deploy-pi.sh` excludes only `node_modules/`. rsync did not treat the symlink as that excluded directory, attempted to replace the Pi’s real `node_modules` directory with the symlink, emitted `could not make way for new symlink: node_modules`, deleted part of the remote dependency tree, and exited 23.

The script stopped before `npm install`, git reset, or service restart, so the running process survived. Recovery was to remove the local symlink, run `npm ci` so `node_modules` was a real excluded directory, rerun the normal deploy, restore Pi dependencies, reset to committed main, restart, and verify health.

## Risk

A common worktree optimization (`node_modules` symlink) can make a deploy unexpectedly destructive to remote dependencies. If later script ordering changes or the failure is ignored, this could leave Hugin unable to restart.

## Suggested acceptance

- Fail preflight when local `node_modules` is a symlink, with a clear remediation message; and/or exclude both `node_modules` and `node_modules/` in a way proven against symlinks.
- Add a regression/shell test covering a symlinked local dependency directory.
- Ensure rsync cannot mutate remote `node_modules` before the dependency-install step.
- Preserve `.env` and other existing deployment exclusions.

Exact incident/recovery is also recorded in repo `STATUS.md` and Munin `projects/hugin/status`.

Magnus-Gille:
## Summary

- Adds a hard preflight in `scripts/deploy-pi.sh` that fails fast (before build/rsync) when the local `node_modules` is a symlink, with an exact remediation message (`rm node_modules && npm ci`).
- Hardens the rsync exclude itself as defense-in-depth: `--exclude='/node_modules'` (anchored, no trailing slash — matches a symlink/dir/file alike, unlike the old trailing-slash form which only matches real directories) plus a `-f 'P /node_modules/**'` protect filter so `--delete` can never touch the destination's `node_modules` regardless of exclude misconfiguration.
- Extracts both into `scripts/lib/node-modules-preflight.sh` (mirroring the existing `claude-config-bootstrap.sh` module pattern) so deploy-pi.sh and its regression test share one source of truth for the rsync args.
- Adds `scripts/lib/node-modules-preflight.test.sh`, wired into CI, covering the exact incident shape.
- Preserves `.env` and all existing exclusions.

## Why

During the 2026-07-12 deploy of PR #186 from a clean worktree, local `node_modules` was a symlink to the owning checkout's real `node_modules` (this fleet's own worktree convention). `deploy-pi.sh`'s old exclude (`node_modules/`, trailing slash) only matches real directories, so the symlink fell through it, got transferred as a plain file, and rsync tried to replace the Pi's real `node_modules` directory with it — partially deleting the remote dependency tree and exiting 23 before `npm install`/git-reset/restart ran. Recovery was manual.

## Test evidence

- Verified actual rsync semantics with a local dir-to-dir experiment (never a real deploy, never ssh): the old trailing-slash exclude reproduces `rsync exit 23` against a symlinked source `node_modules`; the new args exit `0` with the destination's `node_modules` contents and `.env` intact.
- `scripts/lib/node-modules-preflight.test.sh` exercises:
  - `node_modules_preflight_check` passing for missing/real-directory `node_modules` and hard-failing (exact remediation message) for a symlink.
  - The **irony guard**: a real `git worktree add` whose `node_modules` is a symlink back to the owning checkout — the exact shape that caused #187 — is caught by the preflight.
  - A real `rsync -av --delete` run with the production `DEPLOY_RSYNC_EXCLUDE_ARGS` proving the destination's `node_modules` and `.env` survive untouched.
- Confirmed the test fails (`exit 23` assertion mismatch) when temporarily reverted to the old buggy exclude pattern, and passes with the fix — red/green confirmed retroactively via the pre-existing local rsync experiment.
- Full suite: `npm run build` succeeds, `npm test` 1657/1657 pass, both `scripts/lib/*.test.sh` pass.

Closes #187

## Test plan

- [x] `npm run build`
- [x] `npm test` (1657/1657 pass)
- [x] `bash scripts/lib/claude-config-bootstrap.test.sh`
- [x] `bash scripts/lib/node-modules-preflight.test.sh`
- [x] Local rsync dir-to-dir experiment reproducing and fixing the exit-23 symlink bug
- [ ] Cross-model (Codex) PR review

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,3 +27,5 @@ jobs:
       - run: npm test
 
       - run: bash scripts/lib/claude-config-bootstrap.test.sh
+
+      - run: bash scripts/lib/node-modules-preflight.test.sh
--- a/scripts/deploy-pi.sh
+++ b/scripts/deploy-pi.sh
@@ -17,17 +17,20 @@ DEPLOY_USER="${DEPLOY_USER:-magnus}"
 REMOTE="$DEPLOY_USER@$PI_HOST"
 REMOTE_DIR="/home/$DEPLOY_USER/repos/hugin"
 
+# Node_modules preflight + hardened rsync exclusion (issue #187). See
+# scripts/lib/node-modules-preflight.sh for the full incident writeup.
+# shellcheck source=lib/node-modules-preflight.sh
+source "$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/lib/node-modules-preflight.sh"
+
+echo "==> node_modules preflight (issue #187)..."
+node_modules_preflight_check "." || exit 1
+
 echo "==> Building locally..."
 npm run build
 
 echo "==> Syncing to $REMOTE:$REMOTE_DIR..."
 rsync -av --delete \
-  --exclude='node_modules/' \
-  --exclude='.git' \
-  --exclude='.git/' \
-  --exclude='.env' \
-  --exclude='tests/' \
-  --exclude='.DS_Store' \
+  "${DEPLOY_RSYNC_EXCLUDE_ARGS[@]}" \
   ./ "$REMOTE:$REMOTE_DIR/"
 
 echo "==> Installing dependencies on Pi..."
--- a/scripts/lib/node-modules-preflight.sh
+++ b/scripts/lib/node-modules-preflight.sh
@@ -17,13 +17,23 @@
 #      node_modules is a symlink at all — the safest fix is never reaching
 #      rsync with the dangerous shape in the first place.
 #   2. DEPLOY_RSYNC_EXCLUDE_ARGS: the exclude pattern itself, hardened to
-#      match a symlink/dir/file alike ('/node_modules', anchored, no
-#      trailing slash), plus a 'protect' filter rule as defense-in-depth so
-#      --delete can never remove the destination's node_modules contents
-#      even if the exclude were ever bypassed or misconfigured.
+#      match a symlink/dir/file alike ('node_modules', no trailing slash),
+#      plus a 'protect' filter rule as defense-in-depth so --delete can
+#      never remove the destination's node_modules contents even if the
+#      exclude were ever bypassed or misconfigured.
 # Both layers exist because the preflight alone leaves a bypass path (someone
 # could run rsync directly, or the check could be skipped) — the rsync flags
 # must be independently safe.
+#
+# Deliberately UNANCHORED (no leading '/'): an anchored '/node_modules' only
+# matches the transfer root, so a nested node_modules (e.g. a subpackage
+# under skills/) would fall through both the exclude and the protect filter
+# — reintroducing the exact #187 failure mode one directory deeper, and
+# silently regressing behavior vs. the historical unanchored 'node_modules/'
+# pattern. Matching at any depth restores that coverage while still fixing
+# the type bug (Codex review finding, verified with a nested-directory rsync
+# experiment: both a top-level and a nested destination node_modules survive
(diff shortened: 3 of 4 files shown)


Magnus-Gille:
## Codex cross-model review

Ran a real Codex review (`gpt-5.6-sol`, `model_reasoning_effort=high`, via `codex exec --disable code_mode_host` — the default codex-code-mode-host sidecar binary was missing/broken in this environment, the `--disable code_mode_host` flag worked around it cleanly).

**2 findings, both addressed in a follow-up commit (2d72fc9):**

1. **Medium** — the exclude/protect args were anchored to the transfer root (`/node_modules`), so a nested `node_modules` (this repo has `skills/markdown-frontmatter-normalization/package.json`, a real subpackage shape) would fall through both and reintroduce the exact #187 failure one directory deeper — also a silent regression vs. the historical unanchored `node_modules/` pattern. **Fixed:** dropped the anchor (`--exclude='node_modules'` + `-f 'P node_modules/***'`), verified with a local nested-directory rsync experiment (both a top-level and nested destination `node_modules` now survive `--delete`, neither is transferred from source).
2. **Low** — the regression test silently skipped its core coverage if `rsync` was missing, and had no negative control proving the fixture actually reproduces the original exit 23 with the old buggy pattern. **Fixed:** missing `rsync` now fails the test instead of skipping; added a negative-control run with the old `node_modules/` exclude that asserts exit 23 against the same fixture before the hardened args are asserted to exit 0 on it. Also added a nested-node_modules case and a protect-filter-alone sanity case.

No critical findings. Codex independently reproduced the original exit-23 bug and independently confirmed the anchored-rule regression and the fixed nested-directory behavior by running real rsync itself, not just reading the diff.

Full suite after the fix-up: `npm run build` ✓, `npm test` 1657/1657 ✓, both `scripts/lib/*.test.sh` ✓ (28 total bash assertions).

Magnus-Gille:
Closing this PR as **superseded by #189**, which merged to \`main\` at 2026-07-12T22:13:51Z (shortly after this PR was opened) and already closed #187.

## Comparison for the record

Both fixes solve the same root cause (rsync's trailing-slash exclude only matches real directories, letting a worktree's symlinked `node_modules` through) with the same two-layer approach: a hard preflight (`-L node_modules` check before any build/rsync/ssh) plus a hardened, unanchored rsync exclude pattern (matches symlink/dir/file alike, at any depth — so neither version has the nested-`node_modules` anchoring bug that Codex caught in this branch's first draft, since #189 never anchored to begin with).

Differences, noted here in case they're useful for a future hardening pass (not filing a follow-up issue — this is genuinely minor):
- #189 doesn't add rsync's `-f 'P node_modules/***'` protect-filter rule (defense-in-depth against `--delete` even if the exclude pattern itself were ever bypassed/misconfigured). This branch added it.
- This branch's test used a real `git worktree add` to reproduce the exact incident shape, plus a negative control proving the fixture reproduces the original `rsync exit 23` with the old buggy pattern; #189's `scripts/deploy-pi.test.sh` stubs npm/rsync/ssh to assert none of them ran, plus a separate real-rsync check.

Both are legitimate, independently-verified fixes for the same incident — this one just lost the merge race. No action needed; `main` is in a good state via #189.

PR #189, opened 2026-07-12
fix(deploy): reject symlinked node_modules before sync

### LINKED ISSUE #187: deploy-pi: node_modules symlink bypasses rsync exclude and partially deletes remote dependencies

Magnus-Gille:
Closes #187

## Root cause

`rsync --exclude=node_modules/` matches a directory but not a symlink with that basename. A worktree-level `node_modules` symlink therefore entered the `--delete` sync, attempted to replace the Pi's real dependency directory, and partially deleted remote packages before rsync exited 23.

## Fix

- fail locally when `node_modules` is a symlink, before build, rsync, SSH, or any remote mutation
- provide a clear `unlink node_modules && npm ci` remediation
- exclude both the `node_modules` entry and `node_modules/` contents as defence in depth
- retain the existing `.git`, `.env`, `tests/`, and `.DS_Store` exclusions
- add the deploy shell regression to CI

## Validation

- red/green `bash scripts/deploy-pi.test.sh`
  - proves a symlinked source calls none of npm, rsync, or SSH
  - proves all deployment exclusions remain present
  - uses real local rsync to prove a source symlink cannot replace or delete a destination's real `node_modules` directory
- `bash -n scripts/deploy-pi.sh scripts/deploy-pi.test.sh`
- `bash scripts/lib/claude-config-bootstrap.test.sh`
- `npm run build`
- `npm test` — 97 files / 1,657 tests
- `git diff --check`

## Deploy and verification

No production action was taken from this branch. After merge, deploy from a clean `main` checkout with real local dependencies:

```bash
npm ci
./scripts/deploy-pi.sh
```

Then verify on the Pi:

```bash
ssh huginmunin.local 'XDG_RUNTIME_DIR=/run/user/1000 systemctl --user is-active hugin.service && XDG_RUNTIME_DIR=/run/user/1000 systemctl --user is-enabled hugin.service && curl -fsS http://127.0.0.1:3032/health'
```

Expected: service `active`, `enabled`, and health reports `status:"ok"`, `polling:true`, queue depth 0.

## Rollback

`git revert` the merged commit on `main`, then run the same deploy command. There is no data migration or configuration change. Restoring the old deploy script also restores the original symlink risk, so only roll back if the preflight itself causes an unexpected regression.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,4 +26,6 @@ jobs:
 
       - run: npm test
 
+      - run: bash scripts/deploy-pi.test.sh
+
       - run: bash scripts/lib/claude-config-bootstrap.test.sh
--- a/STATUS.md
+++ b/STATUS.md
@@ -1,5 +1,30 @@
 # Hugin — Status
 
+**Latest session:** 2026-07-12 (Codex: deploy source hardening, issue #187)
+**Branch:** `codex/hugin-187-node-modules-symlink` from `main@565f9e6`; PR pending.
+
+## Latest — `node_modules` worktree symlinks rejected before deploy mutation (#187)
+
+- `scripts/deploy-pi.sh` now fails before build, rsync, or SSH when the local
+  `node_modules` entry is a symlink, with remediation to replace it with worktree-local
+  dependencies via `npm ci`.
+- Rsync now excludes both the `node_modules` entry and `node_modules/` contents. This is
+  defence in depth against the incident where a worktree symlink bypassed the trailing-slash
+  exclusion and rsync partially deleted the Pi dependency tree before exiting 23.
+- New `scripts/deploy-pi.test.sh`, wired into CI, proves the preflight invokes none of npm,
+  rsync, or SSH for a symlinked source; checks all prior deployment exclusions remain; and
+  uses real local rsync to prove a source symlink cannot replace or delete a destination's
+  real dependency directory.
+- Validation: red/green shell regression, Bash syntax checks, TypeScript build, full suite
+  (97 files / 1,657 tests), both shell test suites, and `git diff --check` are green.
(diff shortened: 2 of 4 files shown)


Magnus-Gille (commented):
Independent Codex review: no actionable findings. The preflight is ordered before build/rsync/SSH, the entry-level rsync exclusion defends against symlink traversal while preserving existing exclusions, and the regression uses real rsync to prove an existing remote dependency directory survives. Full local suite and GitHub CI are green.

PR #250, opened 2026-07-19
feat: activate authoritative quality corrections

Magnus-Gille:
Refs #231.

## What

Activates native-v2 Quality Receipt correction emission, which PR #244 deliberately left fail-closed pending authoritative attempt identity from #240.

- `POST /v1/delegate/rate` with a `correction` now derives the execution attempt **solely** from accepted authoritative LearningTask evidence inside the schema-validated structured result: `state === "m5-admitted"`, `evidenceAccepted === true`, and all four durable refs (attempt start, prepared dispatch, replay payload, attempt outcome), cross-checked against the rated task id/namespace. Callers cannot provide or infer attempt identity; every missing/mismatched element fail-closes 409 with a specific reason.
- Idempotent retry: an identical correction re-sent after a lost success response returns the first server-owned artifact unchanged (JCS-canonical comparison excluding server-owned `receiptId`/`ratedAt`); genuinely changed successors remain `QualityReceiptConflictError` forks.
- Correction `ratedAt` is server-owned and strictly advances past the predecessor inside the same CAS snapshot (same-millisecond safe).

## Provenance

Carries forward the independently re-reviewed unpushed commit `b32328a` (2026-07-19 session) minus its STATUS.md hunk, per repo convention that STATUS updates ride separate docs PRs. Re-reviewed in full by the grimnir session (Fable) as the PR gate; diff verdict: approve.

## Verification

- `npm run build` (tsc) clean
- `npm test`: 122 files / **1,972 passed**, including 3 new handler tests (derive-from-evidence, lost-response retry preserves first artifact, changed-successor fork rejection), plus new task-store and quality-receipt coverage
- `git diff --check` clean

## Rollout gate

CI is currently **billing-blocked account-wide** (job not started: payment/spending-limit). Do not merge until CI runs green or the owner explicitly waives it. After merge + deploy, the remaining #231 delta is the live dogfood: one real rejection+correction cycle rated end to end against an admitted attempt (gated on the #240 joint smoke).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #241, opened 2026-07-19
Implement authoritative LearningTaskContract accounting and monthly closes


Magnus-Gille:
Filed by: grimnir

## Problem

LearningTaskContract v1 makes Hugin the owner of capture, M5-join, and evaluation-candidate denominators. The current adoption map has requester stamps (#240) and the durable task/outcome registry (#232), but no Hugin-owned implementation for immutable accounting, erasure-safe membership, or authoritative monthly closes. Completing the existing tickets would therefore still leave the continuous-learning measurements unverifiable.

## Scope

- Emit one append-only denominator decision for every eligible Hugin attempt at capture, M5 join, and evaluation, including failures and boundary exclusions when no valid learning record exists.
- Use the contract's owner/counter/month/task/attempt natural keys, strict correction chains, retry/emission events, and closed failure/exclusion taxonomy.
- Issue trusted pre-occurrence synthetic/migration boundary evidence and authoritative Hugin attempt evidence.
- At denominator capture, issue privacy-safe membership tokens and an authoritative denominator-basis proof that binds the superseded record to its exact applicable counters, so erasure cannot silently shrink the denominator.
- Preserve and verify cross-owner membership evidence needed when gille-inference erases joined exposure records.
- Persist/query an authoritative Hugin accounting ledger independently of producer-supplied contract bodies.
- Close each Hugin-owned UTC-month counter within the declared grace window using a signed/authenticated partition and high-water proof, including a legitimate zero-event partition.
- Export validation material needed by the shared Grimnir contract validator without including raw task content.

## Acceptance criteria

- [ ] Every production attempt has exactly one effective capture decision; every capture-eligible M5 macro decision has exactly one effective join decision.
- [ ] Every evaluation candidate has one effective admitted/failed/excluded decision, and admission occurs only after all referenced evidence exists.
- [ ] Failure and exclusion accounting remains complete even when no learning record was emitted.
- [ ] Synthetic and migration exclusions require trusted evidence declared before occurrence and inside the authorized window.
- [ ] Membership tokens and denominator-basis proof bind owner, superseded record, occurrence month, natural-key digest, and the exact applicable counter set.
- [ ] Tombstone validation cannot downgrade an M5 attempt to a non-M5 basis or omit a counter.
- [ ] Corrections are append-only, target one prior natural key, strictly advance time, and cannot fork or cycle.
[...]

Issue #237, opened 2026-07-19
Ingest external Codex and Pi task receipts into the durable learning registry


Magnus-Gille:
Filed by: grimnir

## Problem

Hugin currently has strong provenance for tasks it dispatches, but Magnus's new primary surfaces include direct Codex App, Codex CLI, and Pi work that may never enter Hugin's task lifecycle. Hugin #232 will add the durable task/outcome registry; without an external receipt intake, that registry remains systematically incomplete and success-biased toward Hugin-managed execution.

Importing whole chat transcripts would create weak task boundaries and unnecessary privacy exposure. External work instead needs an explicit, content-governed receipt path.

## Scope

- Implement a supported authenticated intake for versioned external task receipts under `LearningTaskContract v1`.
- Accept at least start/observation and terminal/outcome receipts keyed by stable producer, source-task, task-instance, and receipt IDs.
- Keep execution ownership explicit: an imported receipt is not a Hugin-dispatched task and must never inherit a false Hugin completion claim.
- Reconcile late terminal receipts, reviews, corrections, successor tasks, PRs/commits, and quality receipts into the append-only registry from #232.
- Preserve canonical raw-task identity and separately preserve rendered prompt/configuration identities when supplied.
- Store governed references/hashes rather than copying private prompt, output, transcript, or diff content into content-blind views.
- Record producer coverage state (`covered`, `imported-incomplete`, `late`, `unknown`, or equivalent) and propagate holdout ineligibility when negative coverage cannot be proven.
- Make replay idempotent, concurrent intake lossless, and local spool/retry safe.
- Define abuse controls, authentication, sensitivity, retention, erasure, and unsupported-version behavior.

## Acceptance criteria

- [ ] A direct Codex App, Codex CLI, and Pi task can each appear in the durable registry without being misrepresented as Hugin execution.
- [ ] Start and terminal receipts reconcile to one task instance across retries and duplicate delivery.
- [ ] One issue/branch/PR task is not conflated with unrelated work from the same chat or side conversation.
- [ ] Missing pre-inference exposure proof fails closed for holdout eligibility.
- [ ] Late review/correction receipts update lineage without rewriting prior evidence.
- [ ] Private content remains outside content-blind registry and operational views.
- [ ] Cross-repo contract fixtures cover accepted, duplicate, late, unsupported-version, unauthorized, and incomplete-coverage cases.
- [ ] Real dogfood proves one direct task from each target surface end to end.

## Dependencies

[...]

Issue #233, opened 2026-07-18
Package qualified production candidates into frozen controlled experiments


Magnus-Gille:
Filed by: grimnir

## Problem

Hugin has two working half-loops:

1. ordinary task → evidence/receipt → daily candidate manifest;
2. manually authored Gate-D manifest → matched experiment → evaluation → reviewed promotion.

No production consumer converts a qualified harvested candidate into a frozen, independently verified experiment sample. Consequently no ordinary harvested task has traversed candidate → experiment → reviewed improvement.

## Scope

- Consume durable qualified candidates, not only the rolling `latest.json` view.
- Reconstruct the task through content-governed references.
- Derive or require an independent verifier/rubric and record its version.
- Recheck exposure immediately before freeze and again before execution.
- Freeze immutable task, repository, verifier, model, prompt, harness, and policy identities.
- Create/import an explicit one-axis champion/challenger experiment.
- Preserve holdout, regression, and quarantine semantics.
- Require product rating where mechanical verification is insufficient.
- Produce an auditable reviewed promotion proposal; do not apply production changes automatically.

## Acceptance criteria

- [ ] One real ordinary Hugin task is packaged and run through the controlled experiment ledger.
- [ ] Exposure is rechecked at freeze and execution time.
- [ ] The sample is immutable and reproducible from exact repository/task/configuration references.
- [ ] The experiment changes exactly one declared axis.
- [ ] Mechanical and product-quality outcomes remain distinct.
- [ ] An inconclusive or failed challenger produces durable evidence without promotion.
- [ ] The first end-to-end run and operator decision are documented.

## References

- Hugin #192 supplies a manual real-ticket harness comparison but does not close this automated bridge.
- The daily factory documentation already calls the independent-verifier packager a future prerequisite.

## Non-goals

- Automatic deployment.
- Training model weights.

Magnus-Gille:
Gille evidence-import counterpart: Magnus-Gille/gille-inference#267. This ticket owns candidate freeze and controlled experiment creation; gille owns admission into capability/routing evidence. Tracked by Magnus-Gille/grimnir#88.

Magnus-Gille:
Renumbering note (2026-07-19 clean-history cutover): `gille-inference#267` cited in the scoping comment above now lives at Magnus-Gille/gille-inference#8 ("Import verified Hugin experiment outcomes into the capability ledger"). The old number only resolves in the private-history archive.

Magnus-Gille:
[...]

PR #255, opened 2026-07-20
Add durable append-only task/outcome learning registry (#232)

Magnus-Gille:
Refs #232

## What

Implements the append-only, idempotent task/outcome learning registry (the mechanism, not the accounting) required by hugin#232 and bound by its 2026-07-19 owner-review comment:

- **`src/learning-registry-schema.ts`** — six content-blind record kinds (`submission`, `attempt-reference`, `terminal-outcome`, `publication`, `correction`, `exclusion-adjustment`), each with a discriminated natural key and a content-derived event id (`reg-<hash>`).
- **`src/learning-registry-store.ts`** — Munin-backed store: idempotent natural-key append (duplicate delivery is a no-op; a genuinely different payload at the same natural key is rejected, never overwritten), correction chains (new immutable identity, predecessor never rewritten, forks structurally impossible), erasure-safe exclusion adjustments (target's counter membership preserved, nothing to "resurrect" since content was never stored), and per-partition high-water hash-chain proof issuance/verification.
- **`src/learning-registry-view.ts`** — `buildTaskLifecycleTimeline` joins one task's registry events into a single ordered timeline, resolving corrections to their effective leaf and surfacing exclusion state, while keeping the full audit trail available.
- **`docs/learning-registry.md`** — design decisions, the store-choice justification, and the #232-mechanism vs #241-consumption boundary.

## Design decisions

**Store choice.** The task brief suggested "Munin envelopes like `task-store.ts`, or SQLite like `learning-task-store.ts`" — but there is no SQLite (or any embedded database) anywhere in this repository. `src/learning-task-store.ts` is itself Munin-backed (`createImmutableLearningArtifact` uses `munin.write(..., create_if_absent: true)`), exactly like `src/broker/task-store.ts`. The registry therefore uses Munin envelopes: it sits next to the existing durable LearningTask attempt rows it references (`tasks/<taskId>/learning-attempt-<uuid>...`, `result-structured`), reuses Munin's existing `create_if_absent`/`expected_updated_at` primitives (already used by `BrokerTaskStore.recordAwait` and `writeQualityReceipt`), and avoids adding a second persistence technology with its own backup/restore story. Full rationale in `docs/learning-registry.md`.

**#232-mechanism vs #241-consumption boundary.** Per hugin#241's 2026-07-20 "Boundary clarification" comment: this PR owns the registry's natural keys, membership evidence issued at capture, and the partition/high-water proof primitives (issuance *and* verification). It does not compute monthly closes, does not talk to `gille-inference`, does not ingest external Codex/Pi receipts (#237), and does not package evaluation candidates (#233) — those stay #241/#237/#233's job, consuming what this module issues rather than re-implementing it.

**Idempotency identity excludes `recordedAt`.** `recordedAt` (when the registry durably accepted an event) is store-observed, not caller-asserted — two calls racing to persist the same logical fact legitimately differ there. The natural-key collision check compares everything *except* `recordedAt`, and always returns the actually-persisted (winning) event rather than the caller's locally-built candidate, so a reader is never told two different truths for one natural key. (Caught by a first draft that would have spuriously rejected legitimate retries — see commit message.)

**Partition proofs are never self-certifiable from a subset.** `issuePartitionProof` is the only path to a `"complete"`/`"empty-confirmed"` status, and it always recomputes the chain digest from the authoritative high-water document's actual persisted members — never from whatever subset a caller happened to load. `verifyPartitionProof` re-derives from current store state (rejecting forged digests and, by default, stale high-water marks) rather than trusting the proof body.

**Not wired into the dispatcher yet.** This is a self-contained, independently tested library. Wiring `record*` calls into `src/index.ts`/`src/broker/handlers.ts` is left to #233/#237, which actually need it — this keeps the PR's blast radius off `src/broker/handlers.ts` while #250/#251 are in flight.

## Verification

- `npm test` — **124 test files / 1980 tests passed** (2 new files, 25 new tests: 21 in `tests/learning-registry-store.test.ts`, 4 in `tests/learning-registry-view.test.ts`). No pre-existing test touched or modified.

Tests cover: duplicate-delivery idempotency (including `recordedAt` divergence across a retry); concurrent-writer safety for both a true duplicate-delivery race and distinct events racing in one partition; correction chains (new identity, immutable predecessor, fork rejection, multi-hop leaf resolution, strict time-advance requirement, unknown-predecessor rejection); erasure adjustments (counter membership preserved, target byte-for-byte unchanged); partial-partition ineligibility (missing high-water doc, missing member event); high-water proof issuance and verification (valid, forged digest, forged empty-confirmed, stale); and the lifecycle view joining real attempt-evidence refs across corrections and exclusions.

**M5 dogfood:** ran a concurrency/edge-case review via `mcp__m5__ask` (qwen3-30b-instruct, `delegator_model_id: anthropic/claude-sonnet-5`) against the store implementation. Verified every finding against the actual code: 3 of 4 flagged "bugs" were false alarms — the "fixes" it proposed were already present verbatim (the CAS loop already re-reads before each retry attempt; `bumpHighWater` already skips an already-counted member; `verifyPartitionProof` already recomputes from the current high-water document). The 4th (a claimed TOCTOU in the correction-chain walk) was a mischaracterization of an inherent, harmless eventual-consistency property rather than a bug, so I added a clarifying doc comment instead of a functional change. Noting this plainly since the review's practical yield was low this time — worth knowing for future dogfooding.

CI is billing-blocked account-wide; this merges after green CI or an owner waiver, and after #250/#251 land first per the task brief.

### DIFF
--- a/docs/learning-registry.md
+++ b/docs/learning-registry.md
@@ -1,0 +1,188 @@
+# Durable append-only task/outcome learning registry (#232)
+
+**Status:** mechanism implemented; not yet wired into the dispatcher's normal
+task lifecycle. `src/learning-registry-schema.ts`, `src/learning-registry-store.ts`,
+and `src/learning-registry-view.ts` are a self-contained, tested library. Wiring
+calls into `src/index.ts`/`src/broker/handlers.ts` is deliberately left to the
+tickets that actually need it (#233, #237), to avoid broadening this PR's blast
+radius across in-flight sibling work on `src/broker/handlers.ts` (#250/#251).
+
+## Scope boundary vs #241
+
+Per hugin#241's 2026-07-20 "Boundary clarification" comment:
+
+- **#232 (this module) owns the mechanism** — the append-only registry, its
+  natural keys, membership evidence issued at capture, and the
+  partition/high-water proof primitives, including their issuance.
+- **#241 owns consumption** — period closes, monthly statements, and
+  cross-owner accounting that *verify and consume* #232's proofs. It must not
+  re-implement or re-issue them.
+
+Concretely: this module never computes a monthly close, never talks to
+`gille-inference`, never ingests external Codex/Pi receipts (#237), and never
+packages evaluation candidates (#233). It only appends events, resolves
+correction chains, and issues/verifies partition proofs that #241 will later
+consume as its authoritative "is this period complete" input.
+
+## Store choice: Munin envelopes, not SQLite
+
+The task brief's framing ("Munin envelopes like `task-store.ts`, or SQLite
+like `learning-task-store.ts`") does not match this repository: **there is no
+SQLite (or other embedded database) dependency anywhere in Hugin.**
+`src/learning-task-store.ts` — the "immutable artifacts" module the issue
+points at — is itself Munin-backed (`createImmutableLearningArtifact` calls
+`munin.write(..., create_if_absent: true)`), exactly like `src/broker/task-store.ts`.
+
+The registry therefore uses Munin envelopes, for two reasons:
+
+1. **It sits next to the evidence it references.** The existing durable
+   LearningTask attempt rows this registry points at
+   (`tasks/<taskId>/learning-attempt-<uuid>`, `-prepared`, `-replay`,
+   `-outcome`, and `tasks/<taskId>/result-structured` — see
+   `docs/learning-task-handshake.md`) already live in Munin under the task's
+   own namespace. Registry events live at `tasks/<taskId>/reg-<hash>` in that
+   same namespace, so a reader never crosses a storage boundary to join them.
+2. **Munin already provides the two primitives this registry needs.**
+   `create_if_absent` gives atomic idempotent-create; `expected_updated_at`
+   gives compare-and-swap. Both are used throughout the existing codebase
+   (`BrokerTaskStore.recordAwait`, `BrokerTaskStore.writeQualityReceipt`,
+   `createImmutableLearningArtifact`). Introducing SQLite would mean a second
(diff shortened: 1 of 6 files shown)


Magnus-Gille:
**Review (grimnir session, Fable — PR gate): APPROVE** (after one requested fix, now applied in f4ba4ee).

Independently verified across both rounds:
- CAS/concurrency design read in full: per-iteration re-read, in-loop membership dedup, create-if-absent first write, bounded retry with loud failure — correct. The store-choice correction (Munin envelopes, not SQLite — this repo has none) was right and is properly justified in the PR.
- Idempotency collision semantics: `recordedAt` excluded from natural-key comparison, persisted winner always returned — correct retry behavior.
- My review found one CONFIRMED defect: `issuePartitionProof` only cross-checked the members→store direction, so an event persisted in the crash window before `bumpHighWater` (never redelivered) could be silently omitted from a "complete" proof. Fixed in f4ba4ee: reverse-direction orphan query degrades to `partial` naming the orphans; query truncation also degrades honestly; two regression tests reproduce both paths. Fresh run: **124 files / 1,982 tests passed**.
- Scope discipline confirmed: mechanism only per the #232/#241 boundary — no closes, no external ingestion, no packaging, not wired into the dispatcher.

Merge after #250/#251 once CI can run (billing-blocked). Note for the merge train: nothing here touches production behavior until #233/#237 wire it in.

Magnus-Gille:
Merged under owner-approved CI waiver (option 2, 2026-07-20): Actions is structurally unavailable on this private repo (account billing); full local suite green, independent review (Fable, PR gate) approved; the post-publication public CI run must validate the final tree before any deploy.

PR #256, opened 2026-07-20
Broker bind resilience: retry tailnet race, surface state in /health

### LINKED ISSUE #252: Broker silently absent after reboot: tailnet-IP bind races tailscaled at boot
Filed by: grimnir (2026-07-20 session, observed live after the huginmunin reboot — brokkr#3)

After the unattended ~02:04 reboot, `hugin.service` came up active and /health green, but the Broker (HUGIN_BROKER_HOST=100.97.117.37, port 3035) was NOT listening — every /v1/delegate/* client got connection failures while the service looked healthy. A manual `systemctl --user restart hugin` at ~02:4x, after tailscaled had the IP up, bound 3035 correctly.

Likely mechanism: the user service starts before tailscaled assigns the tailnet address; binding a not-yet-assigned IP fails; the broker's opt-in design degrades to "dispatcher without broker" silently.

Asks:
- Make the failure loud: /health should report broker=absent (it currently reads fully ok), and startup should log the bind failure at error level.
- Fix the race: retry the bind with backoff until the configured address is available (preferred), and/or order the unit after tailscaled/network-online for user scope, and/or bind 0.0.0.0 with auth unchanged if acceptable.
- Regression test for the degraded state being visible.

Impact tonight: blocked the #240 joint live smoke until diagnosed; laptop MCP clients saw "fetch failed" with no server-side signal.

Magnus-Gille:
Closes #252

## What

Fixes the boot-time broker outage observed live the night of 2026-07-20:
`hugin.service` came up active with `GET /health` reading fully-ok, but the
Broker never actually bound `HUGIN_BROKER_HOST` (the Tailscale interface IP,
`100.97.117.37:3035`) because `tailscaled` had not yet assigned the address
when the user-scope service started — `listen()` failed `EADDRNOTAVAIL`, and
the previous one-shot `startBroker(...).catch(err => console.error(...))`
silently degraded to "dispatcher without broker." Every `/v1/delegate/*`
client and the laptop MCP surface saw bare connection failures with no
server-side signal.

- **`src/broker/bind-retry.ts`** (new): `startBrokerWithRetry()` wraps the
  existing single-shot `startBroker()` with bounded exponential backoff on
  transient bind failures, fails fast on permanent ones, and reports every
  state transition live so `/health` is never silently wrong again.
- **`src/index.ts`**: wires the retry wrapper into the existing broker
  startup path; `/health` now includes a `broker` sub-object; shutdown
  cancels a pending retry via `AbortController`.
- **`hugin.service`**: a documentation-only comment explaining why no
  `tailscaled` ordering dependency was added (see Design decisions).
- **`tests/broker/bind-retry.test.ts`** (new): 23 hermetic tests.

## Design decisions

**Health-status semantics.** `/health`'s top-level `status` field is left as
`"ok"` even when the broker is configured-but-down. Rationale: existing
consumers (Heimdall, monitors) may key off `status === "ok"` to mean "the
dispatcher process is up and polling," which remains true independent of the
broker. Flipping that field's meaning risks a breaking change to unknown
consumers for a benefit (broker visibility) that a dedicated field delivers
just as well. Instead, `/health` gains an explicit `broker` sub-object with
a clearly-named `degraded: boolean` (true only when the broker is configured
but not listening — retrying or permanently failed) plus `state`
(`disabled | starting | retrying | listening | failed`), `host`, `port`,
`attempts`, `lastError`, `lastErrorCode`, `boundAt`, `nextRetryAt`, and
`degradedSince`. A monitor that wants broker-specific alerting keys off
`broker.degraded` or `broker.state`, not top-level `status`. This is the
"if changing the top-level risks breaking consumers, add an explicit
sub-object + clearly-named flag" option from the ticket, taken deliberately
rather than by default.

**Retry bounds.** Transient failures (`EADDRNOTAVAIL`) retry with
exponential backoff (2s base, 30s cap, full jitter on the upper half) bounded
by both an attempt cap (1000, effectively a safety net) and a **5-minute
total wall-clock budget**. Rationale: a tailnet address assignment at boot is
typically seconds, not minutes — 5 minutes is generous slack for a slow
`tailscaled` start while still giving up well before "the broker has been
silently down all night" territory. After the budget elapses the dispatcher
keeps running (does not crash-loop) but reports `broker.state = "failed"`
and `broker.degraded = true` — visible via `/health`, not silent. Permanent
errors (`EADDRINUSE` — something else holds the port; `EACCES` — no
permission) stop retrying immediately (no wasted 5-minute wait for something
that will never resolve by waiting). All other error codes are treated as
retryable-but-bounded by the same 5-minute budget rather than added to the
permanent set — a local model review flagged this as "overly permissive"
(suggesting e.g. `EINVAL`/`EAFNOSUPPORT` should also be permanent); I kept
the permanent set narrow and deliberately: misclassifying a recoverable
error as permanent (giving up in one attempt) is worse than a bounded ~5
minutes of futile retrying before reporting `failed` for a truly permanent
but unlisted code. The duration bound already guarantees "does not retry
forever" for every error class, which is what the ticket actually requires.

**Shutdown safety.** `startBrokerWithRetry` accepts an `AbortSignal`. The
dispatcher's shutdown handler aborts it, which (a) short-circuits a pending
backoff sleep immediately and (b) is checked before every new attempt. A
`bind()` already in flight when shutdown fires is *not* abortable (Node's
`listen()` has no cancellation) — if it resolves successfully after
`shuttingDown` was already set, the `.then()` callback in `index.ts` detects
that and closes the newly-bound broker immediately rather than leaking an
open listener into a dying process.

**Unit file.** `hugin.service` already has `After=/Wants=
network-online.target`, which does not guarantee `tailscaled` has assigned
the tailnet address. Per the ticket's own framing, a real ordering fix isn't
reliably expressible here: this is a user-scope unit, and user-scope systemd
cannot reliably order against the system-scope `tailscaled.service`. Rather
than add a directive that looks like a fix but isn't reliable, I added a
comment documenting why the retry (not unit ordering) is the actual fix. No
functional unit-file change.

## Verification

- `npm run build` (tsc): clean, no errors.
- `npm test`: **123 test files / 1978 tests passed** (baseline on
  `origin/main` acd2b00: 122 files / 1955 tests; +1 new test file, +23 new
  tests, 0 regressions).
- New tests are fully hermetic: bind failures/successes are simulated via an
  injectable `bind` function (no real socket races), backoff delay via an
  injectable `sleep`, and time via an injectable `now`. I initially wrote a
  real-socket EADDRINUSE test (two servers racing for the same ephemeral
  port) but dropped it — this sandbox's Node/Express combination showed a
  spurious `listening` callback firing before the async `EADDRINUSE` error
  on a genuinely occupied port, which would have made that test flaky. The
  hermetic injected-bind tests don't depend on that behavior and are the
  right tool here regardless.
- Covered: broker binds on the first attempt; `EADDRNOTAVAIL` retries and
  succeeds once the address "becomes available"; `EADDRINUSE`/`EACCES` stop
  after one attempt (no retry) and report `failed`; bounded give-up under
  both an attempt cap and a duration cap; shutdown cancellation mid-backoff
  and pre-first-attempt; a defensive fix (see below) for a rejecting
  injected `sleep`; and `/health`'s `broker` field for all of
  disabled/starting/retrying/listening/failed, including that
  `disabled` and `failed` are both non-listening but only `failed` is
  `degraded`.

## M5 review (dogfooding)

Ran a bounded review via `mcp__m5__ask` (`qwen3-30b-instruct`,
`delegator_model_id: anthropic/claude-sonnet-5`) against the retry state
machine, asking specifically about shutdown-during-pending-retry,
double-bind, and `/health` staleness during the retry window (the pattern
this session was asked to watch for: genuine catches mixed with confident
false positives).

- First `ask` call (with file-context attachment) errored: "File attachments
  are disabled — none of the configured HOMESERVER_BLIND_CONTEXT_ROOTS
  resolve to a real directory" in this environment. Retried with the code
  inlined in the prompt instead — a second inline attempt also timed out
  once before a shorter prompt succeeded.
- It correctly confirmed shutdown-during-pending-sleep is handled via the
  abort race in `waitOrAbort`.
- Its "no double-bind/orphan risk" verdict was directionally right but
  reasoned incorrectly — it claimed "abort cancels the loop before [a
  broker is assigned]," which isn't true for an in-flight `bind()` (Node's
  `listen()` isn't cancellable, as the code's own comment says). The actual
  safety mechanism is the caller-side `if (shuttingDown) { rb.close() }`
  check in `index.ts`, which it didn't mention. Right answer, wrong/
  incomplete justification — treated as a partial catch, not fully trusted.
- It flagged treating all non-`EADDRINUSE`/`EACCES` codes as
  retryable-until-budget-exhausted as "overly permissive." Evaluated this
  directly (see Design decisions above) and kept the narrow permanent set
  by design rather than acting on it as a bug — a legitimate design question
  it raised, not a defect.
- Independently (not via M5) I found a real gap while re-verifying its
  claims: `waitOrAbort`'s `sleep(ms).then(...)` had no rejection handler,
  so a broken injected `sleep` would produce an unhandled promise rejection,
  violating the function's documented "never throws" contract. Fixed by
  treating a rejected sleep as "proceed immediately" in both the
  signal-present and no-signal branches, with two new regression tests.

CI is billing-blocked account-wide; merge waits for green CI or owner waiver.


Magnus-Gille:
**Review (grimnir session, Fable — PR gate): APPROVE.**

Verified: permanent-set classification (EADDRINUSE/EACCES fail-fast; EADDRNOTAVAIL retried under a 5-minute wall-clock budget with jittered backoff), /health `broker` sub-object with documented status semantics (top-level `ok` preserved for process-liveness consumers; `broker.degraded` is the monitor key), shutdown abort plus the late-bind close guard both present in src/index.ts. Fresh run after merging current main into the branch: **126 files / 2,030 tests** — exactly baseline plus all four merged PRs' additions plus the 23 new bind tests. The agent's self-caught `waitOrAbort` rejection-handler gap has regression coverage.

Merged under the standing owner CI waiver (option 2, 2026-07-20); the post-publication public CI run validates before any deploy.

Issue #27, opened 2026-04-07
Task group sequencing not respected — tasks dispatched out of order


Magnus-Gille:
## Problem

When multiple tasks in a group are submitted at the same time (all with `pending` tag), Hugin dispatches them in arbitrary order instead of respecting the `**Sequence:**` field in the task content.

## Observed behavior

Submitted 5 tasks as group `20260407-192900-munin-quality-fixes` with sequences 1-5. Hugin picked up tasks 2, 3, and 4 **before** task 1 had completed. Each had a prerequisite check that correctly detected the previous task hadn't finished and SKIPPEd — but the ordering was wrong in the first place.

Timeline:
- All 5 tasks written to Munin within ~2 seconds (19:34:57 → 19:36:36)
- Task 2 started at 19:35:32 — before task 1 had run at all
- Task 4 started at 19:35:56 — before task 3
- Task 1 was still `pending` while tasks 2-4 had already completed (as SKIPPED)

## Expected behavior

When tasks share a `**Group:**` field, Hugin should dispatch them in `**Sequence:**` order — only starting sequence N+1 after sequence N has completed (or failed/timed out).

## Workaround

Submit tasks without prerequisite checks and rely on `git rebase origin/main` to pick up prior changes. Works for independent changes in the same repo but breaks for truly dependent tasks.

## Suggested fix

When polling for pending tasks, if a task has a `**Group:**` field:
1. Parse the `**Sequence:**` number
2. Check if any lower-sequence task in the same group is still `pending` or `running`
3. If so, skip this task and pick up the lower-sequence one first

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #259, opened 2026-07-20
fix: tolerantly parse M5 exposure lookup responses (Closes #258)

### LINKED ISSUE #258: Daily exam factory breaks on additive exposure-lookup fields: consumer parses admin response with .strict()
Filed by: grimnir (found live during the 2026-07-20 deploy session; production regression, active nightly)

## Symptom

`hugin-daily-exam-factory.service` (timer-triggered daily candidate factory) failed 2026-07-20 12:05:44 CEST with exit 1. It wrote 2 content-blind candidate records successfully, then failed at the M5 exposure-join step: `M5 task exposure lookup failed (invalid-response)` (`src/learning/m5-task-exposure.ts`). The failure recurs on every timer run until fixed.

## Root cause

The client parses the gateway `POST /admin/task-exposures/lookup` response with `taskExposureCoverageSchema` and `taskExposureLookupResultSchema`, **both `.strict()`** (reject unknown keys). gille-inference PR #27 (canonical exposure identity + coverage epochs), deployed to the M5 gateway the same day, ADDED fields to that response under the **same `schema_version: 1`** — an additive, normally backward-compatible change. Verified against the live gateway (`5d67156`):

- coverage +6 keys: `coverage_epoch_id`, `restart_count`, `total_restart_gap_ms`, `canonical_identity_capture_started_at`, `direct_loopback_traffic`, `exact_match_semantics`
- result +3 keys: `identity_kind`, `unseen_claim_supported`, `includes_legacy_inexact`

`.strict()` rejects all nine → `invalid-response`. `results.min(1)` is satisfied (the server returns a synthetic `seen:false` row), so that is not the cause.

This is a forward-compatibility defect in the **consumer**: a client should not `.strict()`-parse a response it does not own. In a sense the crash is the fail-closed contract working too bluntly — it refused to proceed rather than silently mis-parse — but the correct behavior is graceful forward-compatibility, not a hard exit.

## Impact

Daily learning-candidate exposure enrichment is broken in production nightly. Candidate records still write; only the exposure join fails. Medium severity — no data loss, but the harvest→exposure signal the daily factory exists to produce is unavailable until fixed.

## Fix

**Minimal (unblocks production):** change `taskExposureCoverageSchema` and `taskExposureLookupResultSchema` from `.strict()` to `.passthrough()` (or `.strip()`), keeping every existing required field validated. Add a regression test feeding a response with unknown extra keys and asserting it parses.

**Correct (consumes the new signal):** #27's `unseen_claim_supported` and `identity_kind` are exactly the fail-closed holdout protection this factory should honor — an exposure result with `unseen_claim_supported: false` must NOT be treated as a fresh/negative holdout. Recommend adding these two fields to the result schema (optional/tolerant) and having the exposure join respect `unseen_claim_supported` when deciding holdout eligibility. This turns the incident into the intended cross-repo integration.

## Acceptance

- [ ] Daily factory completes the exposure join when the gateway response includes #27's additive fields.
- [ ] Existing required-field validation unchanged; regression test proves tolerant parsing of unknown keys.
- [ ] (Correct tier) The join reads `unseen_claim_supported` and never claims a task unseen/fresh when it is false.
- [ ] Cross-ref: the LearningTaskContract owner (Magnus-Gille/grimnir) should state that admin lookup responses may add fields within a `schema_version` and consumers MUST ignore unknown fields — filed as a follow-up note on grimnir#88 / the contract doc.

---
Tracked by the program epic: Magnus-Gille/grimnir#88. Drafted with M5 (qwen/gpt-oss dogfood) and owner-verified against the live gateway.

Magnus-Gille:
Closes #258

## What

Production regression: `hugin-daily-exam-factory.service` failed every nightly run starting 2026-07-20 12:05 CEST with `M5 task exposure lookup failed (invalid-response)`. Root cause: `src/learning/m5-task-exposure.ts` parsed the gateway `POST /admin/task-exposures/lookup` response with `.strict()` Zod schemas. `gille-inference` PR #27 (canonical exposure identity + coverage epochs) added 6 coverage keys and 3 result keys under the *same* `schema_version: 1` the same day it deployed — an additive, normally backward-compatible change that `.strict()` rejected outright.

Both fix tiers from the ticket are delivered:

**Tier 1 — forward-compatible parsing**
- `taskExposureCoverageSchema`, `taskExposureLookupResultSchema`, and the top-level `taskExposureLookupResponseSchema` changed from `.strict()` to `.passthrough()`. Additive gateway fields under the same `schema_version` no longer break the client.
- The top-level schema was made tolerant too for consistency, even though the observed incident only added nested fields — documented inline as a deliberate, uniform application of "don't `.strict()`-parse a response you don't own," so the next additive top-level field doesn't cause a repeat incident.
- Every existing required field is still validated exactly as before; `.passthrough()` only relaxes handling of unrecognized keys, it does not touch known-field validation.

**Tier 2 — consume the fail-closed signal**
- `identity_kind` (string) and `unseen_claim_supported` (boolean) are now typed, recognized optional fields on `taskExposureLookupResultSchema` (not just passed through).
- `applySnapshot()` in `src/learning/daily-task-exam-factory.ts` now quarantines a candidate — rather than clearing it to `unseen-covered`/provisional-holdout — whenever `unseen_claim_supported === false`, i.e. a negative match the gateway itself cannot vouch for is never treated as fresh.

**Residual follow-up (documented, not implemented here, per the ticket's "keep it minimal" guidance):** `identity_kind`/`unseen_claim_supported` are used only for the eligibility gate; they are not yet propagated into the persisted `crossClientExposure.match` manifest shape. Separately, grimnir#88 / the LearningTaskContract doc should record that admin lookup responses may add fields within a `schema_version` and consumers must ignore unknown ones (called out in the ticket as a cross-repo note).

## Verification

- `npm test` — **126 test files, 2037 tests, all passing** (includes 2 new tests in `tests/m5-task-exposure.test.ts` and 1 new test in `tests/daily-task-exam-factory.test.ts`).
- New/updated tests:
  - `tolerates #27's additive coverage/result fields under the same schema_version (regression for hugin#258)` — a response with all 9 documented extra keys parses successfully and `identity_kind`/`unseen_claim_supported` come through typed.
  - `still rejects a response missing a required field (does not over-loosen)` — deleting an existing required field (`coverage_complete`) still fails with `invalid-response`.
  - `never clears an unseen result as fresh when the gateway cannot vouch for the claim (unseen_claim_supported: false)` — an otherwise-clean unseen snapshot (complete coverage, in-window, all required lanes) is quarantined instead of sealed as `unseen-covered`, and does **not** pick up `independent-verifier-required`.
- This fix would have prevented the observed 2026-07-20 12:05 failure: the live gateway response (6 extra coverage keys + 3 extra result keys, same `schema_version: 1`) is exactly the shape exercised by the new regression test, and it now parses successfully instead of throwing `invalid-response`.

## Dogfood note

Used `mcp__m5__ask` (qwen3-30b-instruct, `delegator_model_id: anthropic/claude-sonnet-5`) for a bounded independent check that `.passthrough()` doesn't weaken existing required-field validation and that the `unseen_claim_supported` guard placement in `applySnapshot` is correct/sufficient (no downstream path can re-promote a quarantined candidate back to provisional-holdout). Its answers matched independent manual tracing of `applyCrossClientExposure`/`applySnapshot`/`quarantine`; both points verified rather than taken on faith.

### DIFF
--- a/src/learning/daily-task-exam-factory.ts
+++ b/src/learning/daily-task-exam-factory.ts
@@ -532,6 +532,15 @@ function applySnapshot(candidate: DailyExamCandidate, snapshot: TaskExposureSnap
     return;
   }
 
+  // #27's fail-closed signal: when the gateway itself cannot vouch for its
+  // own negative match, that match must never be treated as a fresh/eligible
+  // holdout. Quarantine before evaluating coverage completeness/window/lanes.
+  if (snapshot.result.unseen_claim_supported === false) {
+    candidate.crossClientExposure.evidence = ["m5-unseen-claim-unsupported"];
+    quarantine(candidate, "cross-client-unseen-claim-unsupported");
+    return;
+  }
+
   const createdAtIsIso = isoTimestampSchema.safeParse(candidate.source.taskCreatedAt).success;
   const createdMs = Date.parse(candidate.source.taskCreatedAt);
   const fromMs = Date.parse(snapshot.coverage.from);
--- a/src/learning/m5-task-exposure.ts
+++ b/src/learning/m5-task-exposure.ts
@@ -25,6 +25,11 @@ const sha256Schema = z.string().regex(/^[0-9a-f]{64}$/);
 const isoTimestampSchema = z.string().datetime({ offset: true });
 const metadataIdSchema = z.string().min(1).max(160);
 
+// `.passthrough()` (not `.strict()`): the gateway owns this contract and may
+// add fields under the same `schema_version` (gille-inference#27 added six
+// coverage keys the same day it shipped). A client must not hard-fail on
+// unknown-but-additive fields it does not yet understand (hugin#258); every
+// field the client actually depends on remains required and validated below.
 export const taskExposureCoverageSchema = z.object({
   coverage_complete: z.boolean(),
   from: isoTimestampSchema,
@@ -37,8 +42,12 @@ export const taskExposureCoverageSchema = z.object({
   historical_rows_skipped_inexact: z.number().int().nonnegative(),
   incomplete_before: isoTimestampSchema,
   incomplete_reasons: z.array(z.string().min(1).max(500)).max(64),
-}).strict();
+}).passthrough();
 
+// `identity_kind` and `unseen_claim_supported` are #27's fail-closed holdout
+// signal: they are recognized (typed) optional fields, not merely passed
+// through, so `daily-task-exam-factory.ts` can read them. Any other unknown
+// key (e.g. `includes_legacy_inexact`) is tolerated via `.passthrough()`.
 export const taskExposureLookupResultSchema = z.object({
   fingerprint_sha256: sha256Schema,
   seen: z.boolean(),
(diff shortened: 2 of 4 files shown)


Magnus-Gille:
**Review (grimnir session, Fable — PR gate): APPROVE.** Merging under the standing owner CI waiver — wait, CI now runs: **green on public CI**, so this is a normal review-approved merge.

Independently verified:
- Both tiers present. Tier 1: coverage/result/top-level schemas `.strict()`→`.passthrough()`, required fields still validated; `identity_kind`/`unseen_claim_supported` typed as optional. Tier 2: the eligibility guard quarantines a candidate when `unseen_claim_supported === false`, placed after the `seen===true` early return and before coverage/window/lane checks.
- The guard correctly uses strict `=== false`, so an absent field (old gateway, legacy row) does NOT quarantine — backward-safe, no false quarantines.
- Fresh run: build clean, **126 files / 2,037 tests** (2,034 baseline + 3: the 9-extra-key parse regression that reproduces the exact live-gateway shape, the over-loosen guard (missing required field still `invalid-response`), and the unseen-claim quarantine). The regression test's payload matches the live `5d67156` response, so it would have prevented the 12:05 failure.
- Residual (documented, correctly deferred): propagate the two fields into the persisted `crossClientExposure.match` manifest; and the contract-doc note (admin responses may add fields within a schema_version; consumers must ignore unknowns) belongs in grimnir — I'll carry that.

Closes #258. After merge: redeploy hugin so the nightly timer stops failing.

Issue #232, opened 2026-07-18
Persist a durable append-only task/outcome learning registry


Magnus-Gille:
Filed by: grimnir

## Problem

The daily exam factory is a rolling recent snapshot, not a durable learning registry. It scans completed tasks in a bounded window and replaces `latest.json`. A receipt added after the window does not make the task reappear, and old candidates disappear from the current manifest.

The eligibility path is also success-biased: failed/cancelled tasks, no-ops, publication failures, non-code work, and later human rewrites do not become first-class task/outcome evidence.

## Scope

- Introduce an append-only, idempotent task/outcome registry keyed by stable task and evidence identities.
- Ingest execution lifecycle, repository outcome, product-quality receipts, corrections/successors, exposure state, and configuration identity as separate dimensions.
- Reconcile late receipts and corrected successors without rewriting prior history.
- Represent failures, cancellation, no-change, checkout failure, publication failure, and non-code outcomes explicitly.
- Keep raw/private task material outside content-blind views; use governed references/hashes.
- Define retention, erasure, and supersession semantics.
- Keep `latest.json` as an operational view if useful, but not as the source of learning truth.

## Acceptance criteria

- [ ] A receipt added more than 48 hours after completion updates the durable task record.
- [ ] Historical candidates remain queryable after later daily runs.
- [ ] Failed, cancelled, no-op, and publication-failed tasks can become negative/regression evidence.
- [ ] Repeated attempts and corrected successors retain lineage.
- [ ] Re-ingestion is idempotent and concurrent writers cannot lose evidence.
- [ ] Content/sensitivity and deletion policy are documented and tested.

## Dependencies and links

- Hugin #225 owns durable repository publication recovery.
- Quality Receipt hardening should supply structured corrections and rubrics.
- This registry is an input to the candidate-to-experiment packager, not an automatic promotion mechanism.

Magnus-Gille:
Planning clarification from the LearningTaskContract v1 owner reviews (Grimnir #86 / PR #89):

- The durable registry must preserve authoritative task-attempt references so exposure-negative evidence cannot be attached to another attempt.
- Append-only corrections need new content-derived record IDs plus immutable correction lineage; a same-ID changed payload is invalid.
- Denominator membership must bind the original natural key, occurrence period, counter, and owner so erasure cannot move July evidence into August or relabel joined/direct traffic.
[...]

Issue #216, opened 2026-07-15
Separate execution completion from semantic acceptance


Magnus-Gille:
## Problem

Hugin currently maps executor exit code `0` directly to the terminal lifecycle tag `completed`. That only proves the harness process ended successfully; it does not prove that the requested repository change was produced or that the result passed an independent quality gate.

The July 2026 dogfood run exposed both failure modes:

- an M5/OpenCode attempt exited `0` after operating in the wrong path and produced no repository change;
- a Claude coding task produced a PR that passed the repository test suite but failed independent review because the regression coverage and several implementation details were not acceptable.

Both are useful execution records, but `completed` is easy for operators and downstream automation to misread as semantic acceptance.

## Proposed contract

Keep executor lifecycle and product quality separate:

- `completed` means only that execution finished without an infrastructure/process failure;
- a structured post-run review records `pass | partial | redo | wrong` plus verification outcome and reviewer provenance;
- managed repository tasks explicitly surface `change:none` when no repository delta exists;
- any harvester, promotion gate, or autonomous merge path must require an acceptable review state rather than treating `completed` as sufficient.

This should reuse the existing `tasks/<id>/feedback` rating vocabulary where possible instead of inventing a second quality model.

## Acceptance criteria

- [ ] Public docs define `completed` as execution completion, not semantic success.
- [ ] A managed repository no-op is machine-readable in `result-structured` and cannot be mistaken for an accepted solution.
- [ ] General Hugin tasks can receive structured post-run quality feedback with reviewer and result/diff binding.
- [ ] Daily exam harvesting preserves the review state and does not infer a golden answer from `completed` alone.
- [ ] Autonomous promotion/merge decisions require explicit accepted quality evidence.

## Evidence

- `tasks/20260714t213923z-dogfood-heimdall88` — correct bounded result, independently verified.
- `tasks/20260714t214415z-dogfood-cassette3` — infrastructure failure; no result.
- Munin #170 dogfood sequence — exit-0 no-op and later independently rejected/repaired outputs.

PR #260, opened 2026-07-20
Package qualified production candidates into frozen controlled experiments

Magnus-Gille:
Refs #233

## What

Adds a read-only **packager** that turns already-qualified production task candidates into an immutable, content-blind, single-axis champion/challenger experiment package, then hands it to the existing durable experiment surface to run.

- `src/learning/candidate-packager-schema.ts` (185 lines) — candidate input, matched-task, package, and fail-closed rejection/refusal-reason shapes.
- `src/learning/candidate-packager.ts` (327 lines) — the logic:
  - `qualifyCandidate` — per-candidate predicate over a learning-registry (#232) task lifecycle timeline + a quality receipt (#216): requires a live (untruncated) timeline, an unexcluded/unsuperseded attempt-reference and terminal-outcome, a `completed` outcome, a quality receipt bound to the same task (and attempt, for v2 receipts), and a rating at or above a configurable minimum (default `pass`).
  - `packageExperimentCandidates` — pool-level: splits qualified candidates into a champion arm (matches a caller-supplied incumbent fingerprint) and a challenger arm (every other distinct configuration fingerprint), refuses more than two distinct configurations, refuses insufficient candidates per arm, and verifies the two arms differ on **exactly** the declared axis using the existing `configurationChangedAxes` helper (`src/learning/experiment-schema.ts`). Emits a content-addressed, immutable `ExperimentPackage`.
  - `toExperimentCreateInput` — maps a package onto `learningExperimentCreateSchema` (the existing champion/challenger create contract).
  - `packageAndHandOff` — fetches each candidate task's **current** registry timeline (the freeze-time exposure recheck — nothing is cached from an earlier read) and calls `LearningExperimentStore.create`, which is itself idempotent.
- `tests/candidate-packager.test.ts` (19 tests) — qualification accept/reject cases, one-axis enforcement (rejects a 2-axis delta, rejects an axis mismatch), content-blindness, idempotent re-packaging, fail-closed refusals with structured reasons, and the handoff to a real `LearningExperimentStore` (repeat call returns `reused: true`).

## Design decisions

- **Experiment-surface handoff seam (the one that matters most here):** the existing champion/challenger contract (`src/learning/experiment-schema.ts` + `experiment-store.ts`, from the earlier learning-loop work) already enforces one-axis validation itself (`learningExperimentCreateSchema`'s `superRefine`) and already accepts a fully-formed `LearningConfiguration` supplied by a caller (see `tests/fixtures/learning.ts` — that's exactly how the existing test suite builds configurations today). So the packager **wires directly into `LearningExperimentStore.create`** rather than emitting a package for a future importer to consume — the shapes were compatible, per the ticket's stated preference. `packageAndHandOff` is the seam: it is the only place this module talks to Munin-backed stores, and it never calls anything but `.create()` (a read-then-idempotent-create), never an execution/evaluation/promotion method.
- **Configuration provenance is caller-supplied, not reverse-engineered.** Hugin's own attempt evidence (`origin_config` in `learning-task-handshake.ts`) proves *which* prompt/harness ran, but never carries model identity — the M5 gateway owns that capability truth (AGENTS.md: "Hugin ... must not build a competing capability truth"). Rather than fabricate a `model` identity Hugin can't actually evidence, `PackagerCandidateInput.configuration` accepts an already-fingerprinted `LearningConfiguration` the same way `LearningExperimentStore.create` already does, and the packager's job is to verify the *pool's* coherence (task-type match, exactly-one-axis delta, no more than two distinct configurations) rather than construct configuration identity itself.
- **Content-blindness by construction, not by scrubbing.** A package's `matchedTasks` entries carry only `taskId`/`attemptId`/`arm`/`taskType`/`qualityReceiptId`/coarse `qualityRating` — never a receipt's free-text `ratingReason`, a prompt, or a diff. Verified by a dedicated test that injects a marker string into a receipt's `ratingReason` and asserts it never appears in the serialized package.
- **Idempotency key excludes `qualifiedAt`.** `packageId`/`idempotencyKey` are a JCS-canonical SHA-256 digest over everything except the wall-clock freeze timestamp, so re-packaging the identical qualified set later returns the same `packageId`, and the second `LearningExperimentStore.create` call returns `reused: true` instead of a duplicate experiment.
- **Freeze-time exposure recheck, execution-time recheck out of scope.** `packageAndHandOff` always re-fetches each task's registry timeline live before packaging (there is no earlier cached view to go stale) — that satisfies the freeze-time recheck. Re-checking exposure again at *execution* time is the execution engine's responsibility (explicit non-goal here).
- **Quality-receipt correction chains are the caller's responsibility.** `qualifyCandidate` trusts the receipt it's handed as the ledger's current effective receipt; it does not itself walk `quality-receipt.ts`'s correction chain (`correctsReceiptId` / `foldQualityReceipt`). Documented inline as a known scope boundary — resolving a receipt's correction chain belongs to whoever assembles `PackagerCandidateInput` before calling this module.

## Non-goals preserved

No experiment execution/evaluation engine changes (wires to the existing store instead of reimplementing it), no promotion/routing decisions (gille-inference#8's territory), no candidate generation (registry/harvest side, #232/#241's territory).

## Verification

- `npm test` — **127 files / 2053 tests passing** (baseline on `origin/main`: 126 files / 2034 tests; this PR adds 1 file / 19 tests).

## Dogfooding note

Used `mcp__m5__ask` (qwen3-30b-instruct) for a bounded adversarial review of the qualification predicate and one-axis detection logic. The review's three "critical bugs" all cited code that does not exist in this module (e.g. `task.attempt.supersededBy`, `task.receipt.taskAttemptId`, an `!== [axis]` array-reference check) — a clean instance of the documented confident-false-positive failure mode. I verified each claim against the actual source and found none applicable. My own re-read did surface one real, minor improvement (a missing-timeline case was mislabeled with the same reason code as a genuinely truncated registry read) which is fixed and covered by a new test.

### DIFF
--- a/src/learning/candidate-packager-schema.ts
+++ b/src/learning/candidate-packager-schema.ts
@@ -1,0 +1,181 @@
+/**
+ * Frozen controlled-experiment packager -- contract (#233).
+ *
+ * Turns qualified, already-produced production candidates into an immutable,
+ * content-blind, one-axis champion/challenger experiment package, then hands
+ * the package to the existing durable experiment surface
+ * (`LearningExperimentStore.create`, src/learning/experiment-store.ts) to run.
+ * See candidate-packager.ts for the qualification/packaging logic itself;
+ * this file only owns the shapes.
+ *
+ * Scope boundary: this module SELECTS and FREEZES already-qualified evidence.
+ * It never generates a candidate (that is the registry/harvest side owned by
+ * #232/#241) and never runs, evaluates, promotes, or routes an experiment
+ * (src/learning/experiment-*.ts already owns that; gille-inference#8 owns
+ * admission into capability/routing evidence). It is read-only over the
+ * learning registry -- qualification never mutates a registry row.
+ *
+ * Content-blindness: a package carries only opaque identifiers, versions, and
+ * digests, matching the content-blind evidence the learning registry
+ * (src/learning-registry-schema.ts) and the champion/challenger contract
+ * (src/learning/experiment-schema.ts) already use. It never copies a quality
+ * receipt's free-text `ratingReason`, a task prompt, or a diff -- evidence is
+ * referenced by receiptId/eventId and coarse classification only.
+ *
+ * Configuration provenance: `PackagerCandidateInput.configuration` is a
+ * caller-supplied, already-fingerprinted `LearningConfiguration` -- the exact
+ * shape `LearningExperimentStore.create` already accepts directly from a
+ * human or another system (see tests/fixtures/learning.ts). Hugin's own
+ * attempt evidence (src/learning-task-handshake.ts `origin_config`) proves
+ * *which* prompt/harness ran; it does not carry model identity, which the M5
+ * gateway owns (AGENTS.md: "the M5 gateway owns model selection ... Hugin
+ * preserves validated provenance ... but must not build a competing
+ * capability truth"). This module therefore does not reverse-engineer a
+ * `LearningConfiguration` from raw stamp fields -- it validates one is
+ * already present, self-consistent, and stable across the whole candidate
+ * pool except for the single declared axis.
+ */
+
+import { z } from "zod";
+import { taskTypeSchema } from "../broker/task-type-metadata.js";
+import { qualityReceiptSchema, qualityReceiptV2Schema } from "../quality-receipt.js";
+import {
+  learningChangeAxisSchema,
+  learningConfigurationSchema,
+  learningExperimentGatesSchema,
+} from "./experiment-schema.js";
+
+const slugSchema = z.string().regex(/^[a-z0-9][a-z0-9-]{1,79}$/);
+const sha256Schema = z.string().regex(/^[a-f0-9]{64}$/);
+const packageIdSchema = z.string().regex(/^pkg-[a-f0-9]{64}$/);
+
+export const nativeQualityReceiptSchema = z.union([qualityReceiptSchema, qualityReceiptV2Schema]);
+
+export const packagerArmSchema = z.enum(["champion", "challenger"]);
+export type PackagerArm = z.infer<typeof packagerArmSchema>;
+
+export const packagerQualityRatingSchema = z.enum(["pass", "partial", "redo", "wrong"]);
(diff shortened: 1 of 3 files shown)


Magnus-Gille:
**Review (grimnir session, Fable — PR gate): APPROVE.**

Independently verified against source (not the agent's report):
- **One-axis enforcement** is real and fail-closed: `configurationChangedAxes(champion, challenger)` → refuses `multi-axis-delta` (>1 changed), `declared-axis-mismatch` (0 changed, or changed ≠ requested axis), each with a specific reason code. Reuses the existing `experiment-schema.ts` axis-diff rather than reimplementing. The `changed.length === 0` branch is documented as defensive/unreachable-barring-collision — correct.
- **Content-blindness is structural**: `PackagerMatchedTask` carries only `taskId`/`attemptId`/`arm`/`taskType`/`qualityReceiptId`/`qualityRating` — no `ratingReason`, prompt, or diff. Verified the type has no free-text escape.
- **Idempotency**: package digest excludes `qualifiedAt`, so re-packaging the same qualified set is a no-op and `LearningExperimentStore.create` returns `reused: true`. Deterministic ordering (sorted matched tasks).
- **Read-only over the registry** (qualifies existing evidence; never generates candidates or mutates rows) — respects the #232-mechanism / packaging boundary and the Non-goals (no execution engine, no promotion/routing).
- **Handoff seam**: wired directly to the existing `LearningExperimentStore.create` (shapes were compatible, champion/challenger contract already enforces one-axis at creation) — the ticket's preferred option.
- Fresh run after merging current main (incl. #259): build clean, **127 files / 2,056 tests**.

Residual (correctly deferred, documented): configuration identity is caller-supplied (Hugin doesn't own model-identity truth — that's M5/gille), so the packager consumes identity rather than deriving it; the real cross-repo binding rides gille#5's evidence-identity wiring.

Note for the record: the implementing agent's M5 review round fully hallucinated (3 fabricated bugs with fake line refs even against inline-pasted source) — the agent correctly discarded all three and found the one real issue itself. This is the strongest confirmed instance of the gille#25 verification-fabrication failure mode; the agent being responsible-for-quality caught it.

Closes #233 on merge.

Issue #234, opened 2026-07-18
Propose one-axis experiments from production evidence in read-only mode


Magnus-Gille:
Filed by: grimnir

## Problem

Even after candidates can enter controlled experiments, the system has no read-only reflection/proposal step that identifies recurring failures and proposes a bounded one-axis change. Continuous improvement remains dependent on an operator manually noticing patterns and authoring every experiment.

## Scope

- Read durable task/outcome and experiment evidence.
- Group recurring failure modes by task type and producing configuration.
- Propose a single explicit change axis: model/route, prompt, harness, verifier, or tool policy.
- Cite the evidence and counter-evidence behind each proposal.
- Select regression and holdout samples without exposing private content.
- Emit a reviewable experiment draft, never a production mutation.
- Measure proposal usefulness and false-positive rate before considering more autonomy.

## Acceptance criteria

- [ ] A proposal names one axis, champion/challenger identities, sample selection, verifier, guardrails, and abort criteria.
- [ ] Every claim links to durable evidence rows or task references.
- [ ] The job cannot promote, deploy, rewrite routing, or modify prompts.
- [ ] Duplicate/superseded proposals are detected.
- [ ] At least one proposal is reviewed against a manually authored equivalent.

## Dependencies

- Durable task/outcome registry.
- Candidate-to-experiment packager.
- Immutable prompt/harness/model identity.

Magnus-Gille:
Delivered in PR #261 (merged, CI-green). Read-only proposer: scans registry+receipt evidence, groups by task-type then configuration fingerprint, detects single-axis deltas via configurationChangedAxes, ranks by effectSize × sampleSize × recencyWeight (z-test as a separate confidence caveat, not folded in), and fails closed with typed decline reasons on single-config / multi-axis / insufficient-N / sub-threshold populations. Content-blind; shape-aligned to the #233 packager (proposalToPackageRequest→PackageRequest, matched tasks re-resolved from durable stores by id — proven end-to-end to 'packaged'). 12 new tests. This is the front of the experiment pipeline: propose → package(#233) → run → import(gille#8).

PR #261, opened 2026-07-20
Propose one-axis experiments from production evidence (read-only)

Magnus-Gille:
Refs #234

## What

Adds a strictly read-only proposer that scans qualified production evidence and proposes ranked one-axis champion/challenger experiments, precedeing the #233 candidate packager in the pipeline.

- `src/learning/experiment-proposer-schema.ts` -- contract: `ExperimentProposal`, content-blind `matchedTasks` (reuses the packager's own `PackagerMatchedTask` shape), `ProposalOutcome`, and fail-closed `PopulationDeclineReason` codes.
- `src/learning/experiment-proposer.ts` -- logic: `proposeExperiments` (pure), `proposeExperimentsFromRegistry` (read-only registry wrapper), `proposalToPackageRequest` (maps a proposal onto the #233 packager's `PackageRequest`).
- `tests/experiment-proposer.test.ts` -- 12 tests.

It scans `PackagerCandidateInput` rows (the same shape the #233 packager consumes) plus their registry timelines, groups them by task type then by `configuration.fingerprint`, and for every pair of distinct configurations within a task type checks whether they differ on exactly one axis (via the existing `configurationChangedAxes`). When both arms clear a minimum sample size and the quality-rate delta clears a minimum threshold, it emits a ranked proposal; otherwise it records a concrete decline reason instead of a weak proposal.

It never packages, creates, runs, evaluates, or promotes an experiment -- those stay entirely inside `candidate-packager.ts` and `experiment-store.ts`.

## Design decisions

- **Ranking heuristic**: `score = effectSize * sampleSize * recencyWeight`, where `effectSize` is the absolute pass-rate delta, `sampleSize = min(championSamples, challengerSamples)` (an experiment's power is bound by its thinner arm), and `recencyWeight` is an exponential half-life decay (default 14 days) on the population's most recent evidence. A normal-approximation two-proportion z-test is reported alongside as `evidence.confidence`/`evidence.zScore`, deliberately *not* folded into the score, since it isn't robust to non-independent samples or multiple-comparison inflation across the several populations one run may propose. Documented in the module doc comment.
- **Rating floor differs from the packager on purpose**: the packager floors `qualifyCandidate` at `"pass"` (it only ever freezes already-winning candidates). The proposer floors at `"wrong"` (the lowest rating) by default, because it needs the full outcome spectrum to detect a quality delta between two configurations at all -- flooring at `"pass"` here would make every surviving population's pass rate trivially 1.0. Every *other* `qualifyCandidate` contamination check (timeline completeness, exclusion/erasure, superseded evidence, receipt binding) still applies unchanged.
- **Proposer -> packager shape alignment**: `ExperimentProposal.matchedTasks` reuses `packagerMatchedTaskSchema` verbatim -- content-blind (taskId/attemptId/arm/taskType/qualityReceiptId/qualityRating) references only, no free-text `ratingReason`, no raw task content. `proposalToPackageRequest` maps `(taskType, axis, champion, challenger, matchedTasks)` directly onto `(scope, hypothesis, changeAxis, championFingerprint, gates)`. It deliberately does **not** return `PackagerCandidateInput[]` -- the caller must re-resolve each matched task by id from its own durable stores (registry + quality-receipt ledger) before calling the packager, mirroring the same re-fetch-by-id discipline `packageAndHandOff` already applies to registry timelines rather than trusting a cached view. Exercised end-to-end in the last test (proposal -> `proposalToPackageRequest` -> resolved candidates -> `packageExperimentCandidates` -> `"packaged"`).
- **Champion/challenger assignment**: within a one-axis-differing pair, the group whose evidence started appearing earlier is treated as the incumbent (champion); the later group is the challenger under test. Deterministic tie-break on fingerprint string when timestamps coincide.
- **Content-addressed `proposalId`** excludes every wall-clock-dependent field (`recencyWeight`, `rank.score`, `proposedAt`) so the same underlying evidence pool re-analyzed later keeps the same identity -- this doesn't build a dedup conductor (out of scope here) but gives one a cheap identity to compare against.

## Verification

- `npm run build` -- clean (`tsc`, strict mode).
- `npm test` -- **128 test files passed (128), 2068 tests passed (2068)** (baseline before this change: 127 files / 2056 tests; this PR adds 1 file / 12 tests, no regressions).

Tests cover: a valid one-axis proposal from a seeded two-config population; refusal on multi-axis delta; refusal on insufficient samples in either arm; refusal on a single-configuration population; decline when the quality-rate delta doesn't clear the configured threshold; fail-closed behavior on contaminated evidence (identical to the packager's `qualifyCandidate`); refusal when nothing survives qualification; ranking order across multiple populations (effect size x sample size x recency); content-blindness (no raw `ratingReason` text anywhere in a serialized proposal); `proposalId` stability across reruns at a different wall clock; the read-only registry wrapper; and the proposal -> packager end-to-end shape alignment.

### Dogfooding note (M5)

Ran a bounded `mcp__m5__ask` review (qwen3-30b-instruct) against the literal ranking-heuristic/one-axis-detection source, asking specifically for edge cases in the score formula, the pairwise grouping/champion assignment, and the confidence z-test. All three findings it returned were verified against the actual source and **none held up**:
1. Claimed the recency weight could exceed 1 under clock skew -- but the exact code shown to it already clamps `ageDays` via `Math.max(0, ...)`, which its own answer quoted and then ignored.
2. Claimed a tie-break bug when two fingerprints are equal -- structurally impossible, since the compared fingerprints come from distinct keys of a dedup'd `Map`.
3. Restated the already-documented, intentional sample-size fallback for a degenerate standard error as a "flaw" without identifying an actual defect.

Consistent with this program's prior rounds: M5 fabricated a bug against inline source it was directly shown. Net for this round: not useful as a bug-finder, though verifying its claims did force a second confirmation pass over the clamp and the fingerprint-uniqueness invariant.

### DIFF
--- a/src/learning/experiment-proposer-schema.ts
+++ b/src/learning/experiment-proposer-schema.ts
@@ -1,0 +1,191 @@
+/**
+ * Read-only one-axis experiment proposer -- contract (#234).
+ *
+ * Scans already-qualified production evidence (the same qualified-candidate
+ * shape the #233 packager consumes, `PackagerCandidateInput`) and proposes
+ * which one-axis champion/challenger experiments are worth packaging. It
+ * never packages, creates, runs, evaluates, or promotes anything -- see
+ * experiment-proposer.ts's module doc comment for the read-only guarantee.
+ *
+ * Shape alignment with the #233 packager (candidate-packager-schema.ts):
+ *  - `ExperimentProposal.matchedTasks` reuses `packagerMatchedTaskSchema`
+ *    verbatim -- the exact content-blind (taskId/attemptId/arm/taskType/
+ *    qualityReceiptId/qualityRating) reference shape the packager itself
+ *    freezes into a package. A proposal never carries more than the packager
+ *    would keep.
+ *  - `experimentProposerModule.proposalToPackageRequest` (in
+ *    experiment-proposer.ts) maps a proposal onto `PackageRequest` --
+ *    (task-type, axis, champion, challenger, matched set) becomes
+ *    (scope, hypothesis, changeAxis, championFingerprint) directly. The
+ *    caller still resolves each `matchedTasks` entry back into a full
+ *    `PackagerCandidateInput` (configuration + native quality receipt) from
+ *    its own durable stores before calling the packager -- exactly mirroring
+ *    how `packageAndHandOff` re-reads registry timelines from bare task ids
+ *    rather than trusting a caller-supplied cache.
+ *
+ * Content-blindness: like the packager's `ExperimentPackage`, a proposal
+ * carries only opaque identifiers, versions, digests, counts, and rates. It
+ * never copies a quality receipt's free-text `ratingReason`, a task prompt,
+ * or a diff.
+ */
+
+import { z } from "zod";
+import { taskTypeSchema } from "../broker/task-type-metadata.js";
+import {
+  learningChangeAxisSchema,
+  learningConfigurationSchema,
+} from "./experiment-schema.js";
+import {
+  packagerCandidateInputSchema,
+  packagerMatchedTaskSchema,
+  sha256Schema,
+  type CandidateRejectionReason,
+  type PackagerCandidateInput,
+  type PackagerMatchedTask,
+} from "./candidate-packager-schema.js";
+
+export { packagerCandidateInputSchema, sha256Schema };
+export type { CandidateRejectionReason, PackagerCandidateInput, PackagerMatchedTask };
+
+const slugSchema = z.string().regex(/^[a-z0-9][a-z0-9-]{1,79}$/);
+const proposalIdSchema = z.string().regex(/^prop-[a-f0-9]{64}$/);
+
+export const proposalConfidenceSchema = z.enum(["low", "medium", "high"]);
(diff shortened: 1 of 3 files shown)


Magnus-Gille:
**Review (grimnir session, Fable — PR gate): APPROVE.** Verified against source: ranking is `effectSize × sampleSize × recencyWeight` with the two-proportion z-test kept as a SEPARATE confidence caveat (not folded into the score) — statistically honest; the recency clamp `Math.max(0, …)` is present (the exact code M5 falsely flagged as an overflow bug). Correctly floors qualification at `wrong` (a proposer must see the full outcome spectrum to detect a delta), fail-closed declines with typed reasons (single-config, multi-axis, insufficient-N, sub-threshold), content-blind (`PackagerMatchedTask` only). Shape alignment is the good part: `proposalToPackageRequest`→`PackageRequest` and it deliberately does NOT return `PackagerCandidateInput[]` — callers must re-resolve matched tasks from durable stores by id, proven end-to-end through `packageExperimentCandidates` to `packaged`. Fresh run: build clean, **128 files / 2,068 tests**. Your M5-not-useful verdict confirmed (I re-verified the disputed clamp). Merging on green public CI.

PR #262, opened 2026-07-20
Refs #241: authoritative period closes over #232 registry proofs

Magnus-Gille:
Refs #241

## What

Adds the LearningTaskContract **consumption** layer for Hugin-owned monthly
accounting: `LearningPeriodCloseStore` (`src/learning-period-close.ts`).

`close(period)`:
1. For each accounting counter (`submission`, `attempt-reference`,
   `terminal-outcome`, `publication`), asks #232's `LearningRegistryStore` for
   its partition/high-water proof (`issuePartitionProof`) and independently
   re-verifies it (`verifyPartitionProof({ requireCurrent: true })`).
2. For every counter whose proof certifies, derives `correctedEvents` /
   `excludedEvents` / `erasureAdjustments` / `exclusionAdjustments` by
   resolving each partition member through #232's own
   `buildTaskLifecycleTimeline` — never by re-walking correction/exclusion
   chains itself.
3. Any counter whose proof is ineligible, fails re-verification, or whose
   lifecycle timeline can't be proven complete (truncated pagination, missing
   member) blocks that counter by name. The whole statement then degrades to
   `"partial"` — never a silent full-period claim over a subset.
4. Persists a content-addressed, append-only `PeriodStatement`
   (`close-<32 hex>`) plus a separately CAS-updated "latest" pointer per
   period, so re-closing unchanged state is a no-op and a later
   correction/erasure produces a new, distinct, superseding statement while
   the old one is retained untouched.
5. Optionally joins gille-inference's own owner-issued cross-owner basis via
   an injected `GilleBasisSource` port — referenced or honestly
   `"unavailable"`, never fabricated, and never blocking Hugin's own
   independently-certified counters.

## Design decisions

- **#232-mechanism vs #241-consumption boundary** (per the 2026-07-20
  boundary-clarification comment on #241): this module adds **no new
  registry record kind**, **never re-issues or re-derives** a
  partition/high-water proof (it always asks `LearningRegistryStore` for one,
  then independently re-verifies it), and **never changes #232's natural
  keys**. Correction/exclusion resolution is entirely delegated to #232's own
  `buildTaskLifecycleTimeline` read view rather than reimplemented here.
- **The partial-vs-certified rule**: `status` is `"certified"` iff
  `blockedCounters` is empty, which requires *every* `PRIMARY_ACCOUNTING_COUNTERS`
  entry to have (a) an eligible proof, (b) a passing independent
  re-verification, and (c) a fully provable corrected/excluded count. A
  caller-supplied `counters` subset (mainly useful for focused tests) can
  never reach `"certified"` either — every counter outside the requested set
  is added to `blockedCounters` by name, so the statement always honestly
  degrades to `"partial"` instead of risking a schema-validation crash or a
  silent full claim over an evaluated subset (an edge case a bounded M5
  review pass on the certification predicate prompted me to check directly
  against the code, and it reproduced as a real crash before the fix below).
- **Content addressing excludes call-observed fields**: `closedAt`,
  `supersedes`, and each proof's own `issuedAt` are stripped before hashing,
  so two closes of identical registry state at different wall-clock times
  produce the identical `statementId`.
- **Erasure never shrinks the denominator**: `totalEvents` is always the
  partition's own `highWaterSeq`; `excludedEvents`/`effectiveCount` are
  reported separately and informationally, matching #232's "target's own
  partition membership count never shrinks" invariant.

## Verification

- `npm run build` — clean.
- `npm test` — full suite: **129 files / 2079 tests passing** (baseline was
  128 files / 2068 tests; this PR adds 1 file / 11 tests).
- Dogfooded `mcp__m5__ask` (qwen3-30b-instruct) for a bounded review of the
  certification predicate and partial-statement completeness. It confirmed
  no gap in the four checked properties (no path to a "certified" status
  with an unverified/ineligible/unprovable counter; blocked-handling is
  exhaustive; idempotency/content-addressing has no collision hole) — a
  useful confirmatory pass, though its scope was limited to the pasted
  module and it did not itself catch the `counters`-subset schema-crash edge
  case above, which I found by writing a targeted repro test against the
  actual code.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/src/learning-period-close.ts
+++ b/src/learning-period-close.ts
@@ -1,0 +1,539 @@
+/**
+ * Authoritative monthly (period) close — LearningTaskContract consumption
+ * layer (hugin#241).
+ *
+ * Scope boundary (2026-07-20 "Boundary clarification" on hugin#241, echoed in
+ * the header of src/learning-registry-schema.ts): hugin#232 owns the
+ * *mechanism* — the append-only registry's natural keys, membership evidence
+ * issued at capture, and the partition/high-water proof primitives
+ * (`issuePartitionProof` / `verifyPartitionProof`). This module owns the
+ * *consumption* — period closes, monthly statements, and cross-owner
+ * accounting that VERIFY and CONSUME #232's proofs. It does not:
+ *   - add a new registry record kind or counter,
+ *   - re-issue or re-derive a partition/high-water proof (it always asks
+ *     `LearningRegistryStore` for one and then independently verifies it), or
+ *   - change #232's natural keys.
+ * Correction/exclusion resolution reuses `buildTaskLifecycleTimeline`
+ * (#232's own read view) rather than re-walking correction chains here.
+ *
+ * A period statement is content-addressed and append-only: the same
+ * registry state for a period always produces the same `statementId`, so
+ * re-closing is a no-op. A later correction or erasure changes the derived
+ * corrected/excluded counts (never the underlying partition's own
+ * `highWaterSeq` — see #232's "target's own partition membership count never
+ * shrinks" invariant), which changes the digest and therefore produces a
+ * *new*, distinct statement. The old statement is retained untouched; only a
+ * separate, explicitly mutable "latest" pointer moves.
+ *
+ * Certification is fail-closed: a full-period statement is `"certified"`
+ * only when every constituent counter's partition proof is eligible for
+ * certification (#232's `isEligibleForCertification`) AND independently
+ * re-verifies (`verifyPartitionProof` with `requireCurrent: true`) AND its
+ * derived corrected/excluded counts could be computed without a truncated
+ * read. Any single failure degrades the *whole* statement to `"partial"` and
+ * names exactly which counter/partition blocked certification — never a
+ * silent full-period claim over a subset.
+ */
+
+import { MuninWriteRejectedError, type MuninClient } from "./munin-client.js";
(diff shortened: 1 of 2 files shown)


Magnus-Gille:
**Review (grimnir session, Fable — PR gate): APPROVE.** Verified:
- **Boundary honored** (#232 mechanism vs #241 consumption): consumes `issuePartitionProof`/`verifyPartitionProof({requireCurrent:true})` and `buildTaskLifecycleTimeline` — never re-issues a proof, never re-walks correction chains, no new registry record kind, no natural-key change. Exactly the split I clarified on the issue.
- **Certification is fail-closed at the schema level:** the statement schema refinements make a 'certified' statement with any blocked counter (or missing any counter) structurally unconstructable, and a 'partial' must name ≥1 blocker. So a silent-partial-as-certified is impossible by type, not just by runtime check — the right rigor for an accounting close.
- Content-addressed + append-only with a CAS 'latest' pointer; re-close of unchanged state is a no-op; correction/erasure yields a superseding statement (old retained). Cross-owner basis is an injected port (references gille's basis or honest-unavailable, never fabricates, never blocks Hugin's own certified counters).
- Self-caught bug (counters-subset fully-succeeding crashed on the 'must bind every counter' refinement) now degrades gracefully with named blockers + regression test.
- Fresh run: build clean, **129 files / 2,079 tests**. Merging on green public CI.

Issue #236, opened 2026-07-18
Fail closed or isolate execution after managed checkout failure


Magnus-Gille:
Filed by: grimnir

## Problem

For a managed-repository task, checkout/fetch failure is currently non-fatal: execution may continue without a task branch, and normal post-task finalization then cannot isolate or publish the result. An editing task can leave the managed workspace dirty and influence later work while being recorded as execution-completed plus `checkout-failed`.

Production workspace separation protects the deploy checkout but does not make cross-task contamination acceptable.

## Scope

- For mutation-capable managed tasks, fail closed when the requested clean base/branch cannot be prepared, or switch to a fresh disposable worktree with equivalent provenance.
- Distinguish read-only tasks where degraded execution is explicitly permitted.
- Never run an editing agent in a workspace whose base state is unknown.
- Preserve diagnostic and recovery evidence without masking the executor outcome.
- Clean up disposable worktrees safely after evidence is durable.

## Acceptance criteria

- [ ] Mutation-capable execution cannot start after an unresolved checkout failure in a reusable workspace.
- [ ] Read-only degraded behavior, if retained, is explicit in the contract and result.
- [ ] Tests cover fetch failure, dirty workspace, missing branch, disposable fallback, and cleanup failure.
- [ ] A failed preparation cannot contaminate the next task.
- [ ] Result evidence distinguishes preparation, execution, repository, and publication outcomes.

Magnus-Gille:
Delivered in PR #264 (merged, CI-green). A managed task now runs only once its checkout is verified clean (porcelain+ignored empty) and at the pinned base commit — checked after checkout, before any executor. A mutation-capable task against a contaminated/unverified checkout is REFUSED (dispatch short-circuits, executor never invoked, durable checkout-contaminated state); read-only tasks proceed degraded but skip finalizeTaskBranch so a prior failure's dirty leftovers are never committed under a later task. Durable git-native contamination marker; one gated verified in-place recovery (reset --hard pinned + clean -fdx + re-verify), fail-closed if still dirty. 25 tests.

Issue #225, opened 2026-07-15
Make managed-repository publication failures durable and recoverable


Magnus-Gille:
## Problem

A managed task can finish successfully, produce an exact repository commit, then fail to publish it. Hugin currently marks the task `completed`; the push error exists only in service logs, no PR exists, and an operator cannot discover or resume delivery from `result-structured` alone.

## Live evidence

Task `tasks/20260715t093850z-dogfood-cassette4` on deployed Hugin commit `63ca41574f6e2b1991ced8f3560de955ecc7f14d`:

- correctly resolved `origin/HEAD` to `master` in a repository with no `main`;
- Codex exited 0 and Hugin created exact head `bf7d15f78d9d4c202cd36020769eb0e20a60cca4` from base `4e26895d04ccf48c8e196a05f863ca9b0edaac7c`;
- `git push` failed because GitHub denied `grimnir-bot` write access;
- terminal state was still `completed`, with no machine-readable publication failure;
- the exact commit was later recovered from the Pi and published from an authorized identity as Cassette PR #28, without rerunning paid work.

## Proposed contract

- Persist a structured repository-publication outcome separately from executor outcome: `not-needed | pending | published | failed` plus failure kind, branch, resolved base, exact head, and optional PR URL.
- Surface publication failure in terminal tags/health and friction telemetry; do not make service logs the only evidence.
- Preserve a durable recovery path that republishes the existing exact commit without rerunning the model.
- Add a claim-time repository write-capability check or explicit capability registry where trustworthy; fail before paid execution when missing permission is knowable.

## Acceptance criteria

- [ ] A post-run push/PR failure is machine-readable in `result-structured` and task tags.
- [ ] Exact branch/base/head evidence survives the failure.
- [ ] An authorized operator can retry publication idempotently without rerunning the agent.
- [ ] Known missing repository write capability rejects before paid execution.
- [ ] Tests cover permission denied, push success/PR failure, retry, and completed executor plus failed publication.

Magnus-Gille:
The audit confirms this recovery work is also essential learning evidence: publication failures are valuable negative/regression samples and must not disappear into service logs.

Keep this issue focused on exact-commit delivery recovery. Hugin #232 owns ingestion of the resulting publication outcome into the durable task/outcome registry. Tracked by Magnus-Gille/grimnir#88.

Magnus-Gille:
[...]

PR #264, opened 2026-07-20
Refs #236: fail closed or isolate execution after managed checkout failure

Magnus-Gille:
Refs #236

## What

Hugin runs managed-repository tasks against **reused** working directories — one directory per repo name, shared across many sequential tasks over time, not a fresh checkout per task. `checkoutTaskBranch`'s pre-task fetch/checkout failure was historically non-fatal: the caller logged a warning and executed the agent anyway (`repositoryOutcome: "checkout-failed"`, execution continued). Worse, `git checkout -b` can succeed while silently carrying an earlier task's uncommitted/untracked leftovers onto the new branch instead of conflicting — so "checkout succeeded" was never proof the tree was actually clean. A dirty checkout could then be auto-committed and published by `finalizeTaskBranch` as if it were the *current* task's own output, and the *next* task inherits the same contaminated directory.

This PR adds a pre-execution isolation/verification gate around the managed-checkout seam:

- **Durable contamination marker** — `markCheckoutContaminated` / `readCheckoutContamination` / `clearCheckoutContamination` (`src/task-helpers.ts`) store a git-native record via `git config --local hugin.checkout-contaminated`, scoped to exactly the reused working directory, invisible to `git add -A`/`git diff` (git never tracks its own metadata dir), and durable across dispatcher restarts.
- **Clean-verification gate** — `verifyCleanCheckout(workingDir, expectedCommit)` proves the tree has no uncommitted/untracked/**ignored** state (`git status --porcelain --ignored`) *and* that `HEAD` matches the exact resolved base commit, before any executor runs.
- **`prepareManagedCheckout`** wraps `checkoutTaskBranch` with detection + capability-gated recovery: a mutation-capable task gets exactly one explicit recovery attempt (hard reset to the pinned commit + remove untracked/ignored files) and a re-verify; if still unverified, it fails closed and the marker stays (never cleared on assumption). A read-only task proceeds in explicit degraded mode instead, and `index.ts` skips `finalizeTaskBranch` for it so a prior task's leftover dirty state is never auto-committed/published in the degraded task's name.
- **`index.ts` wiring** — on `refusalReason`, the entire runtime dispatch chain is skipped (no ollama/claude-sdk/codex-spawn/opencode/orchestrator invocation); a failed result is synthesized instead. The model-execution path itself is untouched.
- New `checkout-contaminated` `repositoryOutcome` state distinguishes a gate refusal from the existing `checkout-failed` (fetch/network) outcome.

## Design decisions

**Contamination marking (durable + discoverable).** Stored via `git config --local` inside the repo's own `.git/config` rather than a sidecar file or Munin record — it's naturally scoped to exactly the reused directory (no path-hashing needed), it's invisible to any tree-scanning git command (`add -A`, `diff`, `status`), and it needs no new fs-mocking test infrastructure (stays consistent with the existing `spawn`-based test style in `repo-sync.test.ts`). Detection/marking is **unconditional** — a proven-untrustworthy checkout is recorded regardless of whether the *discovering* task can itself mutate the tree, so the evidence survives for whichever task looks at the directory next. Only the *recovery attempt* is gated on mutation capability.

**Fail-closed vs. isolate.** The issue offered two options: fail closed, or switch to a fresh disposable worktree with equivalent provenance. I chose **fail-closed with one explicit, verified in-place recovery attempt** rather than disposable `git worktree` isolation: recovery here (`git reset --hard <pinned-commit>` + `git clean -fdx`) only ever touches the local working directory — no push, no remote mutation, fully re-derivable from the remote — so it carries none of the external-side-effect risk that made #225's publication recovery operator-only. A disposable worktree would add real complexity (creation, cleanup, provenance-matching against the existing branch-per-task-in-shared-directory model) for no safety benefit over a verified in-place reset. If the one recovery attempt doesn't produce a *verified*-clean tree, the task refuses — never falls through to execution, never assumes success.

**The clean-verification gate.** `checkoutTaskBranch` succeeding is not proof the tree is clean (git doesn't require a clean tree for `checkout -b` unless there's a literal conflict), so verification is a **separate, mandatory** step: `git status --porcelain --ignored` (empty) + `git rev-parse HEAD` == the pinned base commit captured *before* the checkout ran. `--ignored` was added after an M5 review caught that plain `--porcelain` would let gitignored leftovers (e.g. a stale `.env` from a crashed prior task) verify as "clean" while `git clean -fdx`'s `-x` flag would remove them — closing that gap so verification and recovery agree on what "clean" means.

## Verification

- `npm test` — full suite: **133 test files / 2137 tests, all passing** (baseline before this change: 132 files / 2112 tests; +1 file / +25 tests, zero regressions).
- New `tests/checkout-isolation.test.ts` (spawn-mocked, mirrors `repo-sync.test.ts`/#225's style): durable marking + independent read-back, the verification gate (dirty tree, HEAD mismatch, invalid commit, git errors, idempotent-when-clean, gitignored-leftover regression), recovery (success / reset-failure / clean-failure), the full gate flow (skip / fetch-failure mark+refuse / fetch-failure mark+degrade / dirty-recover / dirty-recovery-fails-stays-marked / dirty-degrade-no-recovery-attempted / clean-clears-stale-marker / two-runs-idempotent-both-refuse), and a spy-based contract test proving a refused gate never reaches a stand-in executor while a clean one always does.
- **M5 review** (qwen3-30b-instruct, bounded completeness check on the contamination/verification logic): 1 of ~4 claimed gaps was real and was fixed (gitignored leftovers invisible to plain `git status --porcelain`, added `--ignored`); the rest were noise — it described the agent's own legitimate writes during execution as a "TOCTOU flaw," and its ".git/config corruption defaults to clean" claim is factually wrong (the gate always independently re-verifies via `verifyCleanCheckout` regardless of whether the marker is present, readable, or lost).

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -17,7 +17,7 @@ import {
   type MuninClientConfig,
   type MuninReadResult,
 } from "./munin-client.js";
-import { getFoundBatchEntry, extractTaskId, pickEarliestTask, selectNextTask, checkoutTaskBranch, finalizeTaskBranch, deriveRepositoryOutcome, shouldReapExpiredLease, decideStartupRecovery, decideDeliveryRetry, finalizeTaskCompletion, resolveTaskWorkingDirectory, normalizeRoot, parseBaseBranchOverride, DEFAULT_REPOS_ROOT, MAX_TASK_OUTPUT_TOKENS, MAX_TASK_TIMEOUT_MS, parseBoundedPositiveInt, PUBLICATION_FAILED_TAG } from "./task-helpers.js";
+import { getFoundBatchEntry, extractTaskId, pickEarliestTask, selectNextTask, checkoutTaskBranch, finalizeTaskBranch, deriveRepositoryOutcome, prepareManagedCheckout, shouldReapExpiredLease, decideStartupRecovery, decideDeliveryRetry, finalizeTaskCompletion, resolveTaskWorkingDirectory, normalizeRoot, parseBaseBranchOverride, DEFAULT_REPOS_ROOT, MAX_TASK_OUTPUT_TOKENS, MAX_TASK_TIMEOUT_MS, parseBoundedPositiveInt, PUBLICATION_FAILED_TAG } from "./task-helpers.js";
 import { persistPublicationFailure } from "./publication-recovery.js";
 import { queryAllMuninEntries } from "./munin-pagination.js";
 import { executeSdkTask, type SdkExecutorResult, type SdkExecutorOptions, type SdkTaskConfig, type TaskPermissionProfile } from "./sdk-executor.js";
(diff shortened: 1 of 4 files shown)


Magnus-Gille:
**Review (grimnir session, Fable — PR gate): APPROVE.** Verified: the clean-verification gate (`git status --porcelain --ignored` empty + HEAD==pinned base) runs AFTER checkoutTaskBranch but BEFORE any executor (checkout -b succeeding ≠ clean tree — correct); a **mutation-capable** task against a contaminated/unverified checkout is REFUSED (dispatch short-circuits, executor never invoked), while read-only tasks proceed degraded and skip `finalizeTaskBranch` so a prior task's dirty leftovers can never be auto-committed under a different task's name. Durable git-native marker (.git/config, invisible to add/diff). One gated, verified in-place recovery (reset --hard pinned + clean -fdx + re-verify), fail-closed if still dirty. The `mutationCapable` derivation correctly enumerates each executor's write capability. Fresh run: build clean, **133 files / 2,137 tests**. M5 net-useful (caught the gitignored-leftover --ignored gap). Merging on green public CI.

PR #265, opened 2026-07-20
Ingest external Codex/Pi task receipts into the learning registry

Magnus-Gille:
Refs #237

## What

Adds a narrow, authenticated intake for external task/outcome receipts from `codex_app`, `codex_cli`, and `pi` surfaces — work Magnus completes directly through Codex App/CLI or Pi, without it ever entering Hugin's dispatcher. An admitted receipt is mapped into the durable append-only learning registry (#232) as submission / attempt-reference / terminal-outcome events, via the registry's own natural-key idempotent API — this PR does not reimplement or extend that mechanism beyond one additive, backward-compatible field.

New:
- `src/external-receipt-schema.ts` — versioned, content-blind receipt envelope (`.strict()` throughout; identity/instance fields are bounded opaque tokens, never free text; a fixed `reviewerIndependenceNote` literal).
- `src/external-receipt-signing.ts` — HMAC-SHA256 signing/verification shaped like `task-signing.ts`, with its own producer keystore (`HUGIN_RECEIPT_PRODUCER_KEYS[_FILE]`).
- `src/external-receipt-intake.ts` — `ingestExternalReceipt()`: validate → authenticate → optionally reconcile against a native task → idempotently write registry events. Every rejection carries one of twelve specific, auditable reasons.
- `docs/external-receipt-intake.md` — design writeup, including an honestly-flagged accepted limitation (see below).

Changed (minimal, additive):
- `src/learning-registry-schema.ts` / `-store.ts` — `submissionEventSchema.payload.originComponent` widened from `z.literal("hugin")` to an enum (`hugin | codex_app | codex_cli | pi`); `recordSubmission` gained an optional `originComponent` param defaulting to `"hugin"`. No other #232 behavior changed — every existing native call site is unaffected.
- `src/task-signing.ts` — exported three already-existing helpers (`decodeSecret`, `parseKeyStoreJson`, `keyIdMatchesSubmitter`) so the new signing module reuses them instead of duplicating logic.

## Design decisions

**Authenticated intake.** Verification is always mandatory (no `off`/`warn` policy knob, unlike task submission) — this is a brand-new authenticated-only surface. A receipt's `capacityPrincipal` is only trusted once the signature's `keyId` is shown to resolve to a configured key AND alias back to that exact principal (or a `<principal>-<rotation>` form) — the same anti-impersonation binding `task-signing.ts` uses for submitters. Content-blindness is structural: every envelope object is `.strict()` (unrecognized fields rejected, not stripped) and every identity/instance string is a bounded opaque token with no whitespace, so a transcript/prompt/diff cannot fit through even if someone tried.

**#232 reconciliation / no double-count.** A receipt can optionally assert `instance.reconcilesHuginTaskId` to link to an existing native Hugin task. Intake verifies this against the registry's own submission event (`getEvent` + `originComponent === "hugin"`) before honoring it — an unresolvable or non-native target fails closed (`reconciliation-target-not-found` / `-conflict`). taskId/attemptId are otherwise derived deterministically from `sha256(surface, capacityPrincipal, taskInstanceId)`, which is also what makes re-ingestion of the same receipt a structural no-op and makes two different capacity principals (e.g. two subscriptions) derive **different** tasks rather than ever merging into one "independently confirmed" record — a fixed `reviewerIndependenceNote` literal makes that assertion explicit and tamper-proof in every stored record. A submission's own natural key is unchanged (`{taskId}` only), so an imported submission can never silently overwrite, or be overwritten by, a native one — a genuine conflict throws `RegistryNaturalKeyConflictError` rather than merging.

*Accepted limitation (found by review, documented rather than silently left):* reconciliation proves the target is a genuine native submission, but not that it's the *same* underlying unit of work — native submissions don't carry a `sourceTaskRef` to check against, and adding one would be #232 schema scope creep. Blast radius is bounded: only an operator-provisioned producer key can attempt reconciliation, and a wrong claim can only append an attempt onto an already-existing task's timeline, never fabricate or overwrite its submission/terminal state. See `docs/external-receipt-intake.md`.

**Fail-closed coverage states.** Every admitted receipt is honestly marked `imported` or `imported-late` (7-day threshold from `occurredAt` to Hugin's own server-stamped `receivedAt`) — never claimed as native coverage. An unauthenticated, incomplete, or non-content-blind receipt is rejected before admission with a specific reason (`missing-signature`, `unknown-producer`, `producer-mismatch`, `invalid-signature`, `incomplete-envelope`, `non-content-blind`, `unsupported-schema-version`, `unsupported-contract-version`, `reconciliation-target-not-found`, `reconciliation-target-conflict`, `receipt-id-reused-with-different-content`, `registry-natural-key-conflict`) — never silently admitted or silently stripped.

## Verification

- `npm run build` — clean (tsc, no errors).
- `npm test` — **135 test files / 2157 tests passing** (baseline before this change: 132 files / 2112 tests; this PR adds 3 new test files / 45 new tests, all passing, zero regressions).
- Dogfooded `mcp__m5__ask` (qwen3-30b-instruct) for a bounded admissibility/reconciliation review. One real finding accepted (the reconciliation-identity limitation above, now documented); three fabricated findings rejected after checking against actual source (`canonicalEqual` claimed missing but is defined and JCS-based in `learning-registry-schema.ts`; a coverage-lateness sign claim that was backwards; `reviewerIndependenceNote` claimed schema-unenforced but is already a `z.literal`).

Not merging/deploying — leaving for review per house process.

### DIFF
--- a/docs/external-receipt-intake.md
+++ b/docs/external-receipt-intake.md
@@ -1,0 +1,193 @@
+# External task/outcome receipt intake (#237)
+
+**Status:** mechanism implemented (`src/external-receipt-schema.ts`,
+`src/external-receipt-signing.ts`, `src/external-receipt-intake.ts`); not yet
+wired into an HTTP/MCP surface. Wiring `ingestExternalReceipt` behind an actual
+endpoint (and provisioning real `HUGIN_RECEIPT_PRODUCER_KEYS` for Codex CLI /
+Codex App / Pi) is left to a follow-up so this change stays a self-contained,
+independently testable library, matching how #232 itself shipped.
+
+## Why
+
+Magnus's day-to-day work is moving toward Codex App, Codex CLI, and Pi
+surfaces that complete real tasks **without** ever entering Hugin's
+dispatcher. Without an intake path, the durable learning registry (#232)
+stays systematically incomplete and success-biased toward Hugin-managed
+execution — every task Hugin never saw is invisible to the registry that is
+supposed to be the durable record of what actually happened. Importing whole
+chat transcripts would create weak task boundaries and unnecessary privacy
+exposure, so external work instead gets an explicit, content-governed
+receipt path.
+
+This is the Hugin-side of the pair with `gille-inference#10` (the exposure-
+receipt side); `grimnir#90` is the broader rollout/certification program.
+
+## What a receipt is, and is not
+
+A receipt is a small, versioned, content-blind fact: "surface X, capacity
+principal Y, using provider/model/harness Z, observed/completed task instance
+W, at time T." It is never a transcript, prompt, diff, or output. This is
+enforced structurally in `src/external-receipt-schema.ts`, not just by
+convention:
+
+- Every envelope object is `.strict()` — Zod rejects any unrecognised field
+  outright (e.g. a `transcript` field) rather than stripping it silently.
+- Every identity/instance string is an **opaque token**
+  (`externalReceiptTokenSchema` / `externalReceiptRefTokenSchema`): a bounded
+  character class with no whitespace or control characters, so free text
+  structurally cannot fit through an identity field either.
+- The one required "independence" annotation
+  (`CAPACITY_PRINCIPAL_INDEPENDENCE_NOTE`) is a fixed literal, never
+  caller-supplied prose.
+
+## Admission pipeline
+
+`ingestExternalReceipt` (`src/external-receipt-intake.ts`) runs, in order,
+each step failing closed with one of `EXTERNAL_RECEIPT_REJECTION_REASONS`:
+
+1. **Structural + content-blindness validation.** `externalReceiptEnvelopeSchema.safeParse`.
+   A failure is classified into a specific reason — `non-content-blind` for an
+   unrecognised field or a free-text-shaped token, `unsupported-schema-version`
+   / `unsupported-contract-version` for a version mismatch, `incomplete-envelope`
+   for everything else (missing/malformed required fields) — never one opaque
+   "invalid" bucket.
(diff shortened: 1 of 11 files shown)


Magnus-Gille:
**Review (grimnir session, Fable — PR gate): APPROVE.** Verified:
- **Authenticity mandatory:** HMAC-SHA256 (timingSafeEqual, own producer keystore, keyId→capacityPrincipal binding reusing task-signing.ts) — an unauthenticated/forged receipt is rejected, never admitted.
- **Content-blind by construction:** every receipt schema object is `.strict()` (unknown keys rejected outright) + bounded opaque tokens — raw transcript content structurally cannot pass.
- **No double-count:** `reconcilesHuginTaskId` resolves an asserted native link against a REAL native submission event in #232's registry before honoring it (never on faith); the one #232 change (widen `originComponent` to an enum) is minimal and additive. Ingests via #232's natural-key idempotent API — no mechanism reinvented.
- **Fail-closed:** 12 specific rejection reasons (never a generic 'invalid'), honest imported/imported-late coverage marking, idempotent re-ingest, and `reviewerIndependenceNote` as a `z.literal` so distinct subscriptions always derive SEPARATE tasks (never merged as 'independent' evidence — the exact trap the issue warns about).
- Honest documented limitation (reconciliation proves nativeness, not shared task identity — native submissions carry no sourceTaskRef; closing it is #232 scope) flagged in code + docs, not silently left.
- Fresh run: build clean, **136 files / 2,182 tests**. Merging on green public CI. Pairs with gille#10 (the gille exposure-receipt side, next); rollout program is grimnir#90.

PR #268, opened 2026-07-20
feat(learning): orchestrate the experiment cadence (#266)

Magnus-Gille:
Closes #266

## What

`runExperimentCadenceTick` (`src/learning/experiment-cadence.ts`) chains the already-merged experiment-pipeline pieces on one idempotent, fail-closed tick:

1. **propose** — `experiment-proposer.ts` (#234), read-only, over a caller-supplied evidence pool.
2. **dedupe** — pick the top ranked proposal not already "in flight" for its `(scope, changeAxis, championFingerprint, challengerFingerprint)` natural key, tracked in a cadence-owned durable index.
3. **package** — `candidate-packager.ts` (#233) via `packageAndHandOff`, which itself calls the already-idempotent `LearningExperimentStore.create`.
4. **observe** — read (never mutate) every experiment this cadence has ever packaged; one still `running` is reported, not concluded.
5. **conclude** — an experiment whose evaluator decision left "gathering" (`promotion-ready` or `rejected` — it reached its frozen sample target) gets a best-effort gille outcome-export attempt (#8 contract) and a durable, idempotent REVIEWABLE SUMMARY record.
6. **never promote** — `LearningExperimentStore.promote` is not imported by this module. The reviewable summary is the deliverable until gi#49's autonomy controller exists.

New files:
- `src/learning/experiment-cadence.ts` (688 lines) — the orchestrator.
- `src/learning/experiment-outcome-export.ts` (342 lines) — a Hugin-side mirror of gille-inference#34's `POST /admin/experiments/import` wire contract, a real HTTP client, and a pure bundle builder.
- `src/experiment-cadence-cli.ts` (149 lines) — `npm run experiment:cadence` entrypoint, mirroring the `recover:publication` CLI shape (`--candidates <path>`, `--dry-run`).
- Tests: `tests/experiment-cadence.test.ts` (471 lines, 6 tests), `tests/experiment-outcome-export.test.ts` (234 lines, 11 tests), `tests/experiment-cadence-cli.test.ts` (31 lines, 5 tests).

## Design decisions

**Tick composition.** Each stage is independently try/caught: a proposer error, an index-read failure, a packaging refusal, or a store conflict is recorded in `result.errors`/`result.refusals` and the tick continues with the rest. `runExperimentCadenceTick` only rejects its promise for a genuine programming error, never for a business-rule failure — the CLI decides the process exit code from `result.errors`.

[...]

Issue #266, opened 2026-07-20
Continuous improvement: orchestrate the experiment cadence (propose #234 → package #233 → run → import → surface reviewable adoption)


Magnus-Gille:
Filed by: grimnir (2026-07-20 — continuous-improvement gap).

All the experiment-pipeline PIECES exist and are merged — proposer (#234), packager (#233), the experiment store (run), and gille-side outcome import (gille#8) — but nothing ORCHESTRATES them on a cadence. Today each is a manual/library step. For usage to improve continuously, a scheduler must: read production evidence → propose one-axis experiments (#234) → package qualified ones (#233) → run champion/challenger → import outcomes (gille#8) → surface a reviewable capability/routing proposal. Human approves promotion (the gate stays human).

## Scope
- A scheduled orchestrator (owner-run cadence) chaining the existing pieces, one-axis at a time, with the frozen-experiment discipline #233 enforces.
- Fail-closed + shadowed until the measured gate passes (per the epic's HARVEST_MODE/shadow rule) — no auto-promotion.
- Emit a durable reviewable summary ('experiment X: challenger beat champion on task-type T, n=…, awaiting review').

## Acceptance
- [ ] The cadence produces, from real evidence, a completed experiment with an imported outcome and a reviewable proposal — with zero manual orchestration.
- [ ] Promotion still requires human review.

Tracked by Magnus-Gille/grimnir#88 (continuous-improvement cadence).

PR #271, opened 2026-07-21
Add daily systemd timer for experiment cadence

Magnus-Gille:
Refs #266

## What

Wires `npm run experiment:cadence` (merged in #268) into Hugin's established unit-install mechanism so the daily orchestration tick (propose -> package -> observe -> conclude) runs unattended on the Pi, mirroring `hugin-daily-analysis.timer` / `hugin-daily-exam-factory.timer` exactly.

- `systemd/hugin-experiment-cadence.service` -- oneshot. WorkingDirectory/EnvironmentFile/read-only sandboxing conventions copied from `hugin-daily-analysis.service` (the cadence CLI only reads local files and talks to Munin/gille over the network; it never writes locally, so no `ReadWritePaths` needed). `ExecStart` invokes the built CLI directly with flags, same style as `hugin-daily-exam-factory.service`: `node dist/experiment-cadence-cli.js --candidates <snapshot path>`.
- `systemd/hugin-experiment-cadence.timer` -- `OnCalendar=*-*-* 05:00:00`, `Persistent=true`, `RandomizedDelaySec=300`, `WantedBy=timers.target` -- same shape as the two existing timers.
- `scripts/deploy-pi.sh` -- installs + enables the new pair in the same install/enable block as the exam-factory pair (mkdir, cp, `daemon-reload`, `enable --now`). Also seeds `~/.hugin/experiment-cadence/candidates.json` with `[]` if it doesn't already exist yet, so the unit has something to read on day one.

## Known limitation (flagging for review, not fixing here)

`experiment-cadence-cli.ts`'s own doc comment documents that there is no production bulk assembler yet for the `PackagerCandidateInput[]` evidence pool -- that's explicitly out of #234/#233/#268's scope, deferred to a future ticket. Until that generator exists, the timer's `--candidates` snapshot stays the empty-array placeholder seeded by `deploy-pi.sh`. That's still useful, not a no-op: the tick is idempotent and never promotes, and it still safely observes/concludes any already-running experiments each day -- it just won't propose anything new from an empty candidate pool. A follow-up ticket should point `--candidates` at a real generator.

Also checked whether Hugin's docs maintain a consolidated list of systemd timers to extend (per the ticket's item 4) -- no such list exists (only `docs/daily-exam-factory.md`, which documents that one feature specifically), so no doc change was made.

## Verification

- `npm run build` -- clean.
- `npm test` -- **142 files / 2232 tests passed**, matching the stated baseline exactly.
- `shellcheck scripts/deploy-pi.sh` -- shows the same 12 pre-existing info-level `SC2029`/`SC2016`/`SC1091` findings present on `origin/main` (verified by diffing against the unmodified file); zero new findings introduced by this change.
- `systemd-analyze verify` is unavailable on macOS. Verified unit syntax by diffing the new unit files against their templates (`hugin-daily-analysis.service`, `hugin-daily-exam-factory.timer`) -- identical structure except the intended deltas (Description, ExecStart, TimeoutStartSec, OnCalendar, Unit=) -- and by running the built CLI locally with the exact `ExecStart` arguments (`node dist/experiment-cadence-cli.js --candidates /tmp/candidates-empty.json --dry-run`) to confirm it parses flags and executes cleanly.

Not merged or deployed, per scope.

M5 dogfood round skipped this task (spend discipline).

### DIFF
--- a/scripts/deploy-pi.sh
+++ b/scripts/deploy-pi.sh
@@ -119,13 +119,17 @@ ssh "$REMOTE" "
 
 echo "==> Installing user-level systemd services..."
 ssh "$REMOTE" "
-  mkdir -p ~/.config/systemd/user ~/.hugin/daily-exam-candidates
+  mkdir -p ~/.config/systemd/user ~/.hugin/daily-exam-candidates ~/.hugin/experiment-cadence
+  [ -f ~/.hugin/experiment-cadence/candidates.json ] || echo '[]' > ~/.hugin/experiment-cadence/candidates.json
   cp $REMOTE_DIR/hugin.service ~/.config/systemd/user/hugin.service
   cp $REMOTE_DIR/systemd/hugin-daily-exam-factory.service ~/.config/systemd/user/hugin-daily-exam-factory.service
   cp $REMOTE_DIR/systemd/hugin-daily-exam-factory.timer ~/.config/systemd/user/hugin-daily-exam-factory.timer
+  cp $REMOTE_DIR/systemd/hugin-experiment-cadence.service ~/.config/systemd/user/hugin-experiment-cadence.service
+  cp $REMOTE_DIR/systemd/hugin-experiment-cadence.timer ~/.config/systemd/user/hugin-experiment-cadence.timer
   XDG_RUNTIME_DIR=/run/user/1000 systemctl --user daemon-reload
   XDG_RUNTIME_DIR=/run/user/1000 systemctl --user enable hugin.service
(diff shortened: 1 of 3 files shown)


Magnus-Gille:
**Review (grimnir session, Fable — PR gate): APPROVE.** Units mirror the exam-factory pair exactly (only intended deltas); deploy-pi.sh installs+enables the pair alongside the existing ones and seeds an empty candidates file if absent — making the daily tick a safe idempotent no-op on propose/package while still observing/concluding running experiments (honest ENOENT prevention, prominently flagged). shellcheck: zero NEW findings vs main. Build clean, 142/2,232. CI green. Merging; the candidate-pool generator gap is being filed as the follow-up ticket.

Issue #246, opened 2026-07-19
Attempt exact idempotent recovery after an ambiguous stamped M5 response


Magnus-Gille:
Filed by: grimnir

## Context

The adversarial review of #240 found a non-blocking evidence-coverage gap. If a stamped `/delegate` request reaches Gille but Hugin receives an ambiguous transport failure or HTTP 5xx without a trustworthy learning echo, Hugin correctly records `m5-not-admitted` and discards untrusted output. The exact immutable replay payload and idempotency keys nevertheless make an immediate recovery probe possible. Without it, an authoritative admission join may be recoverable only after process restart, reducing join coverage.

This is a resilience improvement, not a correctness blocker for #240: no model output from the ambiguous response may become trusted evidence.

## Acceptance criteria

- After an eligible stamped attempt gets an ambiguous transport failure or 5xx without an authoritative echo, Hugin performs a bounded immediate recovery request using the byte-identical replay payload and the same task, attempt, request, idempotency, principal, namespace, and stamp identity.
- Recovery accepts only Gille’s authenticated, schema-valid exact-recovery response and preserves the same prepared-dispatch cross-binding and storage-classification checks as startup recovery.
- An authoritative recovered admission echo may improve join/evidence state; output from the ambiguous response is never trusted or harvested.
- A missing, malformed, mismatched, or non-recovery response leaves the truthful failure state unchanged and cannot trigger duplicate inference.
- The retry is bounded and does not interfere with normal 429/503 backpressure behavior.
- Regression tests cover successful recovered admission, mismatched recovery rejection, unavailable recovery, byte identity, and no duplicate/fresh inference acceptance.
- Documentation states that this improves evidence coverage but does not convert ambiguous output into a successful task result.

## Relationship

Follow-up discovered during independent review of #240; it should not block that contract-correctness PR.

Magnus-Gille:
[...]

PR #277, opened 2026-07-22
feat(learning): bounded exact idempotent recovery after ambiguous stamped M5 responses (#246)

Magnus-Gille:
## Summary

Implements #246: after an eligible stamped delegate attempt ends in an ambiguous transport failure or non-backpressure 5xx without an authoritative echo, Hugin performs **one bounded (5 s), storage-validated, fail-closed exact-recovery probe** using the byte-identical replay payload and identical task/attempt/request/idempotency/principal/namespace/stamp identity.

- Recovery accepts only an authenticated, schema-valid, admitted+accepted exact-recovery response; the same prepared-dispatch cross-binding and classification checks as startup recovery apply.
- A recovered admission echo improves join/evidence state only — **output from the ambiguous response is never trusted**; missing/malformed/mismatched responses leave the truthful failure state unchanged; no duplicate or fresh inference is possible.
- 429/503 backpressure classification and exit codes are unchanged.

## Validation

- Red/green TDD; focused 5 files / 95 tests; **full suite 145 files / 2,267 tests green**; TypeScript build; `git diff --check`; staged exact-diff adversarial self-review.
- Independently re-verified by mission control (13/13 recovery tests green in fresh shell) and Claude cross-review passed. One non-blocking review note: a 503 still triggers the (cheap, read-mostly) recovery probe against an already-loaded gateway — defensible, but deserves a code comment.
- Motivating live evidence from the same night: a broker submit timed out at 60 s transport while the server had accepted it (idempotency key bridged it), and a separate 120 s gateway abort produced exactly the `transport-not-admitted` ambiguity this PR recovers. Full evidence: `~/mimir/research/hugin-overnight-2026-07-21/REPORT.md`.

## Deliberately not done

No push/merge/deploy in the implementation session, no live M5 testing, no STATUS.md or deployment-file changes.

Part of the 2026-07-21 overnight R&D mission. Implementation: Codex (gpt-5.6-sol); orchestration + independent verification: Claude.

### DIFF
--- a/docs/learning-task-handshake.md
+++ b/docs/learning-task-handshake.md
@@ -111,6 +111,18 @@ reported as outcome-persistence failure without an `attemptOutcomeRef`; startup
 continues with the ordinary failed task result rather than pointing at an
 unusable row.
 
+After an eligible stamped request gets an ambiguous transport failure or a
+non-backpressure 5xx without an authoritative echo, Hugin also makes one
+immediate, short-timeout probe through this same stored-row validation path.
+The probe reuses the byte-identical classified replay payload and every original
+task, attempt, request, idempotency, principal, namespace, and stamp identity.
+Only an authenticated exact stored-admission recovery may improve the admission
+join. Missing, malformed, mismatched, unavailable, or fresh/non-recovery
+responses leave the original `transport-not-admitted` evidence unchanged.
+This improves evidence coverage only: the ambiguous task remains failed, and
+neither the ambiguous response nor the recovery response supplies trusted task
+output or provenance. Normal 429/503 backpressure never enters this probe path.
+
 This contract does not create the learning registry, choose a model, or promote
 evidence. Those remain Hugin #232 and Gille policy/capture responsibilities.
 
--- a/src/homeserver-executor.ts
+++ b/src/homeserver-executor.ts
@@ -34,6 +34,7 @@ import {
   learningTaskExecutionEvidenceSchema,
   requestStampDigest,
   validateLearningTaskGatewayEcho,
+  validatePreparedLearningTaskOutcome,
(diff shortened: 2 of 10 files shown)


Magnus-Gille:
## Independent review — Codex (gpt-5.6-sol, xhigh reasoning)

Two P2 findings, both **validated against source** and now being fixed on this branch (red/green):

1. **[P2] Recovery can fire after external cancellation** (`src/homeserver-executor.ts` catch path + `src/index.ts` closure). The recovery probe uses an independent 5 s signal and does not consult the task's external abort; operator cancellation/shutdown could still trigger a post-cancel `/delegate` request. Fix: suppress recovery when the external signal is aborted and compose it into the probe so mid-probe cancellation aborts it.

2. **[P2] `AbortSignal.any` requires Node ≥20.3** (`src/munin-client.ts:270`) while `package.json` declares `engines: ">=20"`. On Node 20.0–20.2 every signal-bound Munin read in the recovery path throws before fetching — fail-closed (converted to a failed probe) but silently disables the feature. Fix: feature-detected composition helper with listener cleanup + regression test stubbing `AbortSignal.any` to undefined.

A fix commit with regression tests will be pushed to this branch shortly.

Magnus-Gille:
Both review findings fixed in `1e4d61d` (red/green): external cancellation now suppresses and aborts the recovery probe at all three layers, and `AbortSignal.any` has a feature-detected Node 20.0–20.2 fallback with listener cleanup and a regression test stubbing it to undefined. Codex evidence: focused 65/65, full suite 145 files / 2,271 tests, build, `git diff --check`. Independently re-verified: 65/65 focused tests green in a fresh shell before push.

PR #285, opened 2026-07-22
fix(learning): bridge admitted M5 attempts into cadence

Magnus-Gille:
Tracks #284

## Summary

- captures authenticated direct homeserver/M5 attempts in the native learning registry only after primary terminalization
- independently resolves `GET /ledger/{id}` and requires its evidence-identity hash, task ID, attempt ID, model, and canonical task type to match the admitted attempt
- persists `learning-registry:pending` and reconciles partial submission/attempt/outcome writes idempotently to `captured` or permanent `rejected`
- requires the candidate pool and packager to use an exact-bound, independently reviewed Quality Receipt
- proves the default minimum campaign of three samples per arm creates one proposal
- documents the real authority boundary, recovery behavior, and six-sample minimum

## Safety

- Broker/M5 tasks remain outside managed checkout and retain their existing no-file-tool ceiling
- registry and ledger-read failures happen after terminal result persistence, cannot strand paid work, and never trigger a second inference run
- stale/cross-attempt ledger substitution and missing evidence identities fail closed
- gateway response reads are authenticated, time-bounded, size-bounded, and content-blind

## Cross-repository dependency

This PR intentionally remains draft until [gille-inference #61](https://github.com/Magnus-Gille/gille-inference/issues/61) exposes the authoritative content-blind ledger/attempt binding. Current Gille omits those fields, so Hugin will leave eligible captures pending rather than weaken provenance. After that dependency is deployed, #284 still requires a real deployed cadence tick with durable candidate/proposal counts before closure.

## Verification

- `npm test`: 149 files, 2,305 tests passed on exact head
- `npm run build`
- `bash scripts/deploy-pi.test.sh`
- `git diff --check`

Issue #284, opened 2026-07-22
Bridge managed sampler evidence to M5-admitted learning candidates


Magnus-Gille:
## Problem

The deployed continuous-learning cadence is now observable and healthy, but production cannot create an honestly eligible candidate from any current live task path.

This was verified while attempting the post-deploy dogfood step after PR #283. A successful real cadence tick at `2026-07-22T14:53:13.305Z` durably logged zero candidates. Tracing the current contracts shows that the required evidence cannot coexist on one live attempt.

## Contract mismatch

`src/learning/candidate-pool-assembler.ts` requires every candidate terminal outcome to carry all of:

1. `delegation.modelId`;
2. a canonical coding `delegation.taskType`;
3. `attemptOutcomeRef` resolving to `m5-admitted` + `evidenceAccepted` LearningTaskContract evidence; and
4. an exact-bound effective Quality Receipt.

Today:

- the only live native registry writer is `runHarnessLaneSampledAttempt` at the direct Claude/Codex managed-repository execution seam (`src/index.ts`, issue #274);
- those lane outcomes can carry model/task type and repository evidence, but neither the one-shot nor sampled-OpenCode callback supplies an M5-admitted `attemptOutcomeRef`;
- direct `Runtime: homeserver` / Broker M5 tasks can produce the required admitted LearningTask evidence, but Broker explicitly has `tool_policy:none` and no worktree, the dispatcher skips managed checkout for homeserver tasks, and that path is not recorded through the native harness-lane registry writer; and
- external receipt intake is still library-only and its terminal event does not supply the native LearningTask attempt evidence required by the assembler.

Therefore a managed coding task and an M5-admitted candidate are currently disjoint sets. An operator cannot fix this by choosing a better task document or rating a result.

There is a second expectation mismatch: the proposer defaults to three samples per arm, so one dogfood task cannot advance `candidate -> proposal -> experiment`; at least six comparable eligible attempts across exactly one changed configuration axis are required.

## Impact

- The production candidate assembler remains empty even after the cadence service itself is fixed.
- The documented/current handoff instruction to submit one eligible managed task and then exercise candidate -> experiment is not executable.
- Repeated dogfood tasks would spend model/runtime capacity without being able to close the loop.

## Required design decision

Choose one authoritative bridge; do not weaken provenance checks merely to make the pool non-empty. Likely options include:

- add equivalent admitted-attempt evidence to the managed sampler/harness path; or
[...]

PR #286, opened 2026-07-22
fix(learning): normalize Gille evidence hash wire prefix

Magnus-Gille:
## Summary

- accept Gille’s documented `sha256:<digest>` evidence-identity wire format
- validate the prefix strictly, then normalize to Hugin’s canonical bare digest
- keep malformed, legacy, unprefixed, and cross-row responses fail-closed

## Live evidence

After PR #285 and gille-inference #62 were deployed, production task `mcp-m5-cde28030357d13354aa31890` completed with authenticated `m5-admitted` evidence and an exact ledger ID, but remained `learning-registry:pending`. Its durable log showed Hugin rejected the binding response because the deployed Gille endpoint correctly returned the documented `sha256:` prefix.

The pending tag is recoverable; deployment of this patch should capture the existing terminal task without rerunning inference.

## Verification

- red test reproduced the live prefix rejection
- focused integration suite: 3 files / 19 tests passed
- full suite: 149 files / 2,306 tests passed
- TypeScript build passed
- `git diff --check` passed

Tracks #284.

### DIFF
--- a/src/m5-ledger-attempt-binding.ts
+++ b/src/m5-ledger-attempt-binding.ts
@@ -1,12 +1,17 @@
 import { z } from "zod";
 import type { HomeserverGatewayConfig } from "./homeserver-executor.js";
 
-const sha256Schema = z.string().regex(/^[a-f0-9]{64}$/);
+const wireSha256Schema = z.string()
+  .regex(/^sha256:[a-f0-9]{64}$/)
+  .transform((value) => value.slice("sha256:".length));
 
 /** Content-blind join returned by Gille's authenticated GET /ledger/{id}. */
 export const m5LedgerAttemptBindingSchema = z.object({
   id: z.string().min(1),
-  evidenceIdentityHash: sha256Schema,
+  // Gille's ledger contract uses the explicit `sha256:` wire prefix. Hugin's
+  // durable schemas use the canonical bare digest, so normalize only after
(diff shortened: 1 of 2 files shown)


Magnus-Gille:
Review gate complete at exact head `5fdb6b71609b031e1c87a22c580a99e56495d102`.

Independent Codex subagent: **PASS**, no findings. It verified the strict `sha256:<64 lowercase hex>` input contract, canonical bare-digest output type, downstream type safety, focused/full tests, build, and diff check. Reviewer identified itself as GPT-5 Codex.

Cross-model review: M5 `qwen3-coder-next-80b`. Its four proposed findings were validated and declined:

- `fetchM5LedgerAttemptBinding` uses `safeParse` and converts all schema failures to the controlled `M5 ledger binding response did not match...` error;
- unprefixed and null legacy values are deliberately fail-closed, so the asserted rejection paths are correct;
- `resolves.toEqual(binding)` already proves the normalized output is the bare digest;
- the anchored lowercase-hex regex already rejects uppercase/non-hex payloads, so an extra example would be optional coverage rather than a defect.

Live production evidence independently confirms the endpoint returned `sha256:5872...f5e4`, `learningTaskBinding: bound`, and matching ledger/task/attempt/model/task-type identities. No actionable findings remain.

PR #287, opened 2026-07-22
fix(homeserver): forward task model pins

Magnus-Gille:
## What changed

- map the task document `Model` field to the M5 `/delegate` request's existing `modelId` pin
- keep gateway-selected routing unchanged when `Model` is omitted
- route dispatcher construction through a focused, testable mapper
- update the continuous-learning operator docs now that gille-inference #61 is deployed

## Why

Issue #284 requires an honest two-arm production campaign. The parser already accepted `Model` and the executor already supported `modelId`, but the dispatcher dropped the value while constructing direct homeserver tasks. Both campaign arms therefore collapsed onto whichever model M5 selected.

This preserves the authority boundary: direct Broker/M5 tasks still use `tool_policy:none`, receive no managed checkout, and cannot claim repository edits.

## Verification

- red test: focused suite failed because the dispatcher mapper/model pin was absent
- green focused suite: 34 tests passed
- full suite: 149 files / 2,307 tests passed
- `npm run build`
- `git diff --check`

Closes the remaining production-campaign blocker in #284; the issue itself should remain open until the deployed six-sample campaign yields durable nonzero cadence/proposal counts and proves idempotency.

### DIFF
--- a/docs/continuous-learning-loop.md
+++ b/docs/continuous-learning-loop.md
@@ -1,9 +1,9 @@
 # Continuous Hugin/M5 learning loop
 
-**Status:** production cadence active; admitted-attempt candidate capture is
-implemented fail-closed but awaits the authoritative Gille ledger binding
-tracked in [gille-inference #61](https://github.com/Magnus-Gille/gille-inference/issues/61)
-(2026-07-22).
+**Status:** production cadence and admitted-attempt candidate capture are
+active. The authoritative Gille ledger binding tracked in
+[gille-inference #61](https://github.com/Magnus-Gille/gille-inference/issues/61)
+was deployed on 2026-07-22.
 
 ## Goal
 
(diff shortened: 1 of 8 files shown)


Magnus-Gille:
Review record for exact head `af0d8919eb020e976884ccd5ef06308bdfc61630`:

- Independent Codex subagent initially found a blocking authority-boundary issue: whole-document `Model` parsing allowed prompt/Broker JSON text to influence routing.
- Fixed by adding prompt-bounded, line-anchored metadata parsing in `src/task-document-metadata.ts`, using it from production `parseTask`, and adding generic prompt plus serialized canonical Broker regressions.
- Independent subagent re-review: **PASS**, no remaining actionable findings. Reviewer reported OpenAI Codex, GPT-5 family (exact serving identifier unavailable).
- Independent M5 review on `qwen3-coder-next-80b`: **PASS**, no actionable findings after the fix.

Verification at this head: focused dispatcher + Broker task-store + homeserver suites 127/127; full suite 149 files / 2,310 tests; TypeScript build; `git diff --check`.

PR #293, opened 2026-07-22
fix(broker): bound list and submit latency

### LINKED ISSUE #278: broker: /v1/delegate/list and submit can exceed 60s while dispatcher is idle
## Evidence (2026-07-21 overnight R&D, verified live)

- `hugin_list` (MCP → `/v1/delegate/list`) timed out at 60 s transport while SSH loopback `/health` showed the service green: polling, queue depth 0, host load 0.24, broker `listening`.
- One `hugin_submit` also timed out at 60 s; the server HAD accepted it — an idempotent retry with the same key returned `reused_idempotency: true` (contract worked, but the ambiguity window is real).

The cost is inside broker request handling — most plausibly the Munin scan behind list (canonical history discovery caps at 1,000 candidates, status reads at concurrency 25) rather than host or dispatcher overload.

## Suggested direction

- Budget/paginate/cache the list path so a bounded response always returns before the transport deadline; report truncation honestly.
- Keep submit fast-ack (persist minimal, defer enrichment) so its ambiguity window shrinks.

Full context: overnight report `~/mimir/research/hugin-overnight-2026-07-21/REPORT.md`, findings 1.

Magnus-Gille:
Closes #278

## What changed

- batch canonical status reads after the two bounded Munin discovery queries instead of issuing one serialized read per task
- make fresh submit a single atomic `create_if_absent` write and acknowledge only a confirmed `created` result
- recover deterministic idempotent retries from the persisted envelope after an atomic conflict, while preserving principal isolation and collision detection
- retain the in-memory reservation throughout durable conflict validation
- document the load-bearing batching and atomic-create contracts

## Why

The Munin client serializes and spaces requests. Broker list could therefore perform hundreds of individual reads and exceed 60 seconds, while fresh submit performed a preflight read plus a write and could queue behind list traffic. This change bounds list round trips via the existing batch API and removes the submit preflight race without weakening durability.

## Verification

- `npm run build`
- focused Broker suite: 5 files, 152 tests passed
- final focused suite after race adjustment: 3 files, 117 tests passed
- full suite: 152 files, 2343 tests passed
- `bash scripts/lib/claude-config-bootstrap.test.sh`
- `git diff --check`

### DIFF
--- a/docs/mcp-durable-m5-lifecycle.md
+++ b/docs/mcp-durable-m5-lifecycle.md
@@ -24,13 +24,24 @@ Retries compare the normalized behavior payload stored at that namespace;
 rotating the MCP session ID does not create a collision. The same key and
 payload returns the original task, including after Broker restart. The same key
 with a different behavior payload is rejected. An in-process reservation closes
-the concurrent-submit race before the first Munin write completes. Both the
+the concurrent-submit race before the first Munin write completes. The fresh
+path uses one atomic `create_if_absent` write rather than a preflight read plus
+write. If that create loses to an existing durable task, the Broker reads and
+compares the stored envelope before returning reuse or collision. Both the
 durable identity and that reservation are scoped by authenticated principal.
 
 Await and list are principal-isolated. A Broker key cannot discover or read
 another principal's canonical results. Rate remains owner-only for Broker
 tasks; the same authenticated endpoint can also review an ordinary terminal
 Hugin task, which has no Broker owner.
+
+Each of list discovery's two unioned channels walks a bounded 1,000-candidate
+Munin history budget. The response reports `truncated: true` whenever either
+budget or an ambiguous timestamp boundary may omit matches. Discovered status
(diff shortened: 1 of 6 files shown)


Magnus-Gille:
Independent exact-head review gate passed at `1fed929eab08b1a2a7f4fce840cb7fee4913f44e`.

- GPT-5.6 Terra: PASS; verified atomic-create acknowledgement, conflict recovery, restart reuse/collision, principal scoping, batched reads, union discovery, and truncation semantics. Full suite: 2,343 tests.
- Codex GPT-5 family: PASS; verified in-process/cross-process race logic, Munin batch cardinality/order/missing-entry behavior, principal filtering, and documentation/diff hygiene. Full suite: 2,343 tests.

No actionable findings. Non-blocking observation: no dedicated simultaneous two-Broker-process race test; reviewers judged atomic Munin semantics plus conflict-path coverage adequate.

PR #304, opened 2026-07-23
Persist hash-bound scheduler workload snapshots

Magnus-Gille:
## Summary

- derive a conservative aggregate workload snapshot from the complete pending/running pages already fetched by the claim loop
- bind the canonical snapshot digest into the scheduler prediction and persist both create-only after the accepted claim
- preserve lookup-free disabled/truncated behavior, memoize estimates across workload and SEJF observations, and keep all scheduling behavior unchanged
- document explicit incompleteness for running clocks and unenumerated lifecycle buckets

## Verification

- red/green focused scheduler workload/evidence/store tests
- `npm test` (157 files, 2,388 tests)
- `npm run build`
- `bash scripts/deploy-pi.test.sh`
- `git diff --check`

Closes no issue; advances #282.

Issue #282, opened 2026-07-22
scheduling: wire S1-validated shadow-lane experiments (bounded-SEJF, class re-triage) and design work-minute admission control


Magnus-Gille:
## Background (2026-07-21 overnight R&D — simulator-validated)

The discrete-event simulator (8,800 runs, M/M/1-validated to ≤2.15% error; source + report in `~/mimir/research/hugin-overnight-2026-07-21/sched-sim/` and `05-sched-sim-results.md`) established:

- **Complete-window deterministic FIFO is validated** as champion (historical newest-50 policy claimed 0/500 aged tasks — structural starvation). Hardening: PR #276.
- **Bounded SEJF is NOT a default** — its promised ≥25% mean-slowdown gain appeared in 1/16 cells and degrades with estimate noise and near saturation. Zero starvation-bound violations though: it is legitimate SHADOW material.
- **Class re-triage (low→normal at half-SLA)** preserved high-class p95 in 8/8 cells while cutting low-class >2h waits 5.1–86.3% — shadow-worthy with authenticated urgency.
- **At ρ≥1 no ordering helps** — admission control (work-minutes, not task count) is the real lever; "ordering only redistributes delay".

## Proposed slices

1. Shadow-mode logging in the claim path: record what FIFO chose AND what bounded-SEJF (T=30 min) / re-triage would have chosen, with predicted vs realized service — zero behavior change, evidence only. Feed the champion/challenger cadence.
2. Duration-estimate telemetry (needed by any SEJF challenger; estimate noise was the gain-killer in simulation).
3. Design note for work-minute admission: queued work-minutes as the capacity signal, explicit rejection/deferral instead of hidden delay.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Magnus-Gille:
Design slice completed and deployed.

- PR #294 squash-merged as b1e1024374d69a509a056cb34c342d7c2c5e4903.
- Exact final head passed independent Codex GPT-5-family and GPT-5.6 Terra reviews.
- PR CI and exact post-merge CI passed.
- Production marker is the exact merge commit; service, Broker, queue observability, model hosts, and the in-service Codex sandbox are healthy.
- FIFO remains the only live scheduler. No SEJF or work rejection was enabled.
- Follow-up #295 tracks authenticated signing of Group/Sequence and future scheduling metadata.

[...]

PR #305, opened 2026-07-23
Persist hash-bound scheduler workload snapshots

Magnus-Gille:
## Summary

- derive a conservative aggregate workload snapshot from the complete pending/running pages already fetched by the claim loop
- bind the canonical snapshot digest into the scheduler prediction and require that exact prediction before create-only snapshot persistence
- keep `orch-v1` pending rows visible under explicitly incomplete `otherNonterminal`
- preserve lookup-free disabled/truncated behavior, memoize estimates across workload and SEJF observations, and keep scheduling behavior unchanged

## Review fixes

GPT-5.6 Terra found two P2s on the original head: unverified prediction/snapshot storage binding and omitted `orch-v1` pending work. Both are fixed with regression tests in the exact head on this PR.

## Verification

- red/green focused scheduler workload/evidence/store tests (32 passing)
- `npm test` (157 files, 2,389 tests)
- `npm run build`
- `git diff --check`

Supersedes #304, whose PR head remained stuck on its superseded commit after the branch ref advanced. Advances #282.

### DIFF
--- a/docs/scheduling-shadow-lanes.md
+++ b/docs/scheduling-shadow-lanes.md
@@ -426,10 +426,12 @@ capacity, Hugin must say so truthfully.
 
 ### Executable workload snapshot primitive
 
-`src/scheduler-workload.ts` now defines the strict, behavior-neutral
-`schedulerWorkloadSnapshotSchema` and pure
-`buildSchedulerWorkloadSnapshot(...)` builder. It does not query Munin, write
-evidence, defer a task, reject a task, or influence the claim candidate.
+`src/scheduler-workload.ts` defines the strict, behavior-neutral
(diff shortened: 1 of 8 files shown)


Magnus-Gille:
Independent exact-head review: **PASS** by GPT-5.6 Terra at `f3747330befdb8ea56718087afb6dd52062236d9`.

Both prior P2s are fully resolved. Terra independently re-ran the 32 focused tests, TypeScript build, and diff check. Residual risk is limited to pending/running pages being observational rather than one atomic cross-lifecycle snapshot; the evidence is explicitly non-enforcing and fails incomplete on conflicting state.

PR #321, opened 2026-07-25
fix: slim the Heimdall descriptor and make status/deploy honest

Magnus-Gille:
## Summary

The owner's complaint: "loads of alarms, not sure if they are correct, and
the info display is often bloated and not very helpful." Hugin's
`/heimdall.json` was **7,754 bytes live** (measured 2026-07-25) — 12x
munin-memory's 648-byte reference, one of only two services that publish
their own descriptor. Three issues, all addressed:

1. **Bulk data inlined, not described.** The descriptor embedded the full
   `#164` learning-loop capability-evidence table (per task-type/model
   attempts, quality %, maturity, recommendation) on every ~60s Heimdall
   poll. Moved to a periodic push (`src/heimdall-report.ts`) onto Heimdall's
   **existing** `POST /api/panels` store (heimdall issue #57) — the same
   pattern already used in production by mimir
   (`mimir/src/heimdall-report.ts`) and gille-inference's
   `scripts/post-*-panel.ts`. **No heimdall-side change required** — the
   ingest endpoint, auth, and per-service page rendering already exist and
   already serve other producers. Env-gated
   (`HEIMDALL_HUB_URL` + `HEIMDALL_FLEET_TOKEN`) and fail-soft; a documented
   no-op until those are set in Hugin's `.env` on the Pi.
2. **`deploy` block was a structure of nulls.** Now populates
   `deploy.deployed_commit` from the same `.deployed-commit` stamp file
   Heimdall's `drift.js` already reads authoritatively (rsync deploys
   exclude `.git/`, so this stamp is the one honest source). Deliberately
   **omits** `latest_commit`/`drift`/`deployed_at` rather than nulling them —
   Hugin doesn't try to recompute Heimdall's own drift authority (matches
   munin-memory's pattern, the 648B reference).
3. **`status` was a hardcoded `"pass"` literal.** Now derived from real
   dispatcher signals (poll-loop alive, broker bind degraded, blocked-task
   count) via `deriveDescriptorStatus`. Uses the existing (currently inert)
   health+json `output` field to express **"ran fine, N findings"** — e.g.
   `warn` + `"3 blocked task(s) awaiting attention"` — rather than a flat
   binary, while staying inside the current `pass`/`warn`/`fail` schema. This
   is deliberately the shape the parallel Heimdall state-remodeling work
   seems to be heading toward; see "Coordinated change" below.

## Before / after (measured, not estimated)

- Live production descriptor (`ssh huginmunin.local sqlite3 ...`), before this
  change: **7,754 bytes**.
- Local equivalent of the new route, same inputs a running Hugin would supply
  (`health` callback wired, no `.deployed-commit` in a dev checkout): **687
  bytes** — smaller than munin-memory's 648B reference plus its 2 owned
  `plugin`/`view` panel declarations. With a real `.deployed-commit` stamp
  present: **804 bytes**.
- That's an **~11x reduction**, into the same order of magnitude as the
  648-byte reference, not a 12x outlier.

## What moved behind an endpoint vs. removed

- **Moved, not removed**: the capability-evidence table, trial-gate table,
  route-policy status, durable-handoffs stat, and learning-experiments table
  (`buildLearningLoopHealthPanels`, unchanged) now go to Heimdall's panel
  store via push instead of being embedded in the descriptor. The underlying
  evidence computation (`learning-loop-health.ts`) is untouched; only the
  transport changed.
- **Declared, not embedded**: the two `hugin-tasks`/`hugin-history`
  `plugin`/`view` panels stay in the descriptor's `panels` array — they were
  already thin declarations (id/plugin/view/label, no data) and are the model
  for what a descriptor should contain.

## Was Hugin's `status` claim actually correct?

No — it was a static literal, not derived from anything, so it couldn't be
"wrong" in a factual sense but also couldn't be trusted as a signal. It
happened to coincide with "the process is running" most of the time, which
masked the real gap: nothing tied it to broker health or blocked-task
backlog. Separately, Heimdall's "Deploy drift: hugin" alert is computed
**entirely independently** (`heimdall/src/drift.js`, reading the same
`.deployed-commit` stamp + `git ls-remote` on origin/main) — it never reads
Hugin's descriptor at all today, so there was no direct contradiction between
`status: "pass"` and that alert, just two disconnected claims that read as
inconsistent on the dashboard. This PR doesn't fold deploy-drift into
`status` (that would duplicate Heimdall's own drift authority); it makes
`status` honestly reflect dispatcher health, and separately gives the
`deploy` block a real, non-null fact Heimdall's UI can render.

## Coordinated change (flagged, not assumed)

You mentioned a parallel Heimdall change reworking service state into
healthy / completed-with-findings / failed-to-run rather than flat
pass/fail. This PR's `deriveDescriptorStatus` already produces exactly that
shape (`status: "warn"` + a populated `output` string describing the
finding) **using the current schema** — `output` is an existing but
currently-unrendered health+json field in Heimdall's contract
(`src/contract/schema.js` `normalizeDescriptor`), so this should slot into
that work without a producer-side change once Heimdall starts rendering it.
If the new model ends up wanting a distinct third status value or a
structured findings array instead of a string, that would be a
heimdall-side schema addition — flagging it here rather than guessing at
its shape.

## Test plan

- [x] Red first: `tests/heimdall-descriptor.test.ts` (new status/deploy/size
      assertions) and the new `tests/heimdall-report.test.ts` written before
      implementation, confirmed failing for the right reason (module not
      found / old hardcoded-pass / old panel-embedding behavior).
- [x] `npm test` — **159 test files, 2,416 tests, all passing** (real output,
      not asserted).
- [x] `npx tsc --noEmit` and `npm run build` — clean, no errors.
- [x] `bash scripts/deploy-pi.test.sh` — unaffected, all assertions pass
      (this PR does not touch deployment scripts).
- [x] Live descriptor size re-measured via a local Express instance running
      the actual compiled route (not hand-computed).

## Unverified / not done here

- `HEIMDALL_HUB_URL` / `HEIMDALL_FLEET_TOKEN` are not configured in Hugin's
  `.env` on the Pi (out of scope — "do not restart live services"). The
  panel reporter is a safe no-op until an operator sets them; panels will
  keep showing whatever Heimdall already has stored (or nothing) until then.
  Flagging this as the one follow-up needed for the push path to actually
  activate in production.
- Did not attempt to independently recompute deploy drift inside Hugin
  (e.g. `git ls-remote`) — kept that authority solely with Heimdall's
  `drift.js`, per the "don't build a competing capability truth" principle
  already established elsewhere in this codebase for M5 capability evidence.
- Did not modify `heimdall` or `grimnir` — no coordinated change needed for
  this PR's mechanism (push endpoint already exists), only the optional
  `output`-rendering / richer-status-model item flagged above.

### DIFF
--- a/src/heimdall-descriptor.ts
+++ b/src/heimdall-descriptor.ts
@@ -3,9 +3,22 @@
  *
  * Returned verbatim by GET /heimdall.json — no auth required, same posture as /health.
  * Schema: https://heimdall.gille.ai/schema/service/v1
+ *
+ * DESCRIBE, DON'T EMBED (#317-follow-up). This file's job is to say what
+ * exists and how to fetch it — identity, status, real deploy facts, and panel
+ * *declarations* — not to ship bulk data. The learning-loop capability/product
+ * evidence (issue #164) used to be inlined here as full table rows on every
+ * ~60s Heimdall poll; it is now pushed on a much slower cadence to Heimdall's
+ * `POST /api/panels` store (src/heimdall-report.ts) and rendered on Hugin's
+ * service page independently of this descriptor. That is the established,
+ * already-deployed "panel data endpoint fetched on demand" pattern (heimdall
+ * issue #57 — see mimir's src/heimdall-report.ts for a live producer example),
+ * not a new contract.
  */
 
 import type { Application } from "express";
+import { readFileSync } from "node:fs";
+import * as path from "node:path";
 import {
   buildLearningLoopPanels,
   computeCapabilityPlane,
@@ -25,16 +38,33 @@ export const HEIMDALL_DESCRIPTOR = {
     criticality: "normal",
   },
   kind: "http-service",
+  // Base/fallback status when no live health input is wired (e.g. the bare
+  // `registerHeimdallDescriptorRoute(app)` call in tests). The real server
+  // always supplies `opts.health` so the wire response reflects live state —
+  // see deriveDescriptorStatus below. This literal is deliberately NOT what
+  // ships from the running service.
   status: "pass",
   version: "0.1.0",
+  // Only facts Hugin genuinely knows about its own runtime placement.
+  // `deployed_commit` is added per-request from the `.deployed-commit` stamp
+  // (same file Heimdall's drift.js reads authoritatively) when present — see
+  // readDeployedCommit. `latest_commit`/`drift`/`deployed_at` are intentionally
+  // OMITTED rather than published as null: Hugin cannot know origin/main's
+  // latest commit without duplicating Heimdall's own drift computation, and a
+  // structure of nulls reads as "failed to determine" when it should read as
+  // "not this service's job to say" (munin-memory's descriptor — the 648-byte
+  // reference — follows the same omission pattern).
   deploy: {
     host: "huginmunin",
+    systemd_unit: "hugin",
     platform: "bare-metal",
   },
   metrics: [],
   // Tier-1 services own their panels: Heimdall's static known-panels fallback is
   // not consulted once /heimdall.json exists, so the task views must be declared
   // here (#135). Rendered live by Heimdall's plugins/hugin.js from the Munin DB.
+  // These ARE thin declarations (id/plugin/view/label) — no data — unlike the
+  // learning-loop typed panels, which are pushed, not declared here.
   panels: [
     { id: "hugin-tasks", plugin: "hugin", view: "tasks", label: "Tasks", refresh: 60, fullWidth: true },
     { id: "hugin-history", plugin: "hugin", view: "history", label: "Task history", refresh: 120, fullWidth: true },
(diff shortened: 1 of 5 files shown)


Magnus-Gille:
## Review gate: **APPROVE — with a hard deploy prerequisite** (grimnir orchestrator)

Good work, and the right architectural call: the panel data **moved** to Heimdall's already-existing `POST /api/panels` store rather than being deleted, requiring zero heimdall-side changes. I verified that store is genuinely in production use — `brokkr` (4 panels), `m5-inference` (13), and `mimir` (3) all push to it successfully today. So this follows an established pattern, not a new one.

**Size, independently confirmed at source:** live descriptor was **7,754 bytes**; munin-memory's, the lean comparator, is **648**. 687–804 bytes is the right order of magnitude.

Also good:  is populated from the same  stamp Heimdall's  reads authoritatively, while // are **omitted rather than nulled** — that was the actual defect (a struct of nulls cannot be distinguished from "failed to determine"), and omission is the correct fix. Deriving  from real dispatcher signals instead of a hardcoded  is a genuine improvement in honesty.

## ⛔ Do not deploy this until two env vars are set

This is blocking for **deployment**, not for merge. Verified on the Pi (presence-only checks):

```
HEIMDALL_HUB_URL      -> 0 matches in hugin's .env
HEIMDALL_FLEET_TOKEN  -> 0 matches in hugin's .env
```

The PR notes the pusher is "a safe no-op until an operator sets them." That is true of the *pusher* but not of the *outcome*: the descriptor no longer carries the capability-evidence, trial-gate, route-policy, durable-handoffs, or learning-experiments data, and nothing else will publish it. Deploying as-is would make that evidence **disappear from the dashboard rather than move**, which is the opposite of the intent and would be read as "the data broke."

The established location for this config is a per-service env file — brokkr keeps its `HEIMDALL_HUB_URL` in `/home/magnus/.config/brokkr/env`. Hugin needs the equivalent before this ships.

**Merging now** so the work is not left dangling, with the deploy gate recorded here and carried to the owner. Whoever deploys Hugin next must set both vars first and then confirm a non-empty `panels` row for `service='hugin'` in Heimdall's DB.

Tests: 159 files / 2416 passing; typecheck and build clean; red/green followed on both new test files.

Magnus-Gille:
*(Correction — one paragraph above lost its inline code spans to shell interpretation when I posted it. Restating it intact; nothing else in that review changes, and the deploy prerequisite stands exactly as written.)*

Also good: `deploy.deployed_commit` is populated from the same `.deployed-commit` stamp that Heimdall's `drift.js` reads authoritatively, while `latest_commit` / `drift` / `deployed_at` are **omitted rather than nulled**.

That was the actual defect worth fixing. A struct of nulls cannot be distinguished from "failed to determine" by any consumer, so publishing one is strictly worse than publishing nothing — omission is the correct fix, and it also avoids Hugin duplicating Heimdall's drift authority with a second, weaker answer.

Deriving `status` from real dispatcher signals (poll-loop alive, broker degraded, blocked-task count) instead of a hardcoded `"pass"` literal is a genuine improvement in honesty. A hardcoded pass is not false so much as untrustworthy — nothing tied it to reality, so it carried no information.

PR #338, opened 2026-07-27
W4.3: strict Hugin autonomous config adapters

### LINKED ISSUE #331: W4.3: Restrict autonomous config adapters to Hugin prompt, harness/tool-policy, and macro routing
## Purpose

Make the narrowly allowed Hugin-owned targets concrete and enforceable: prompt, harness/tool-policy, and macro-routing configuration only.

## Scope

- Introduce a small versioned configuration adapter/store for the W0-approved Hugin targets, using the canonical source defined by Grimnir #170.
- Migrate the existing Orin macro-route constants into that bounded adapter without broadening behavior: only homeserver public/internal `classify`/`extract` remain eligible; private, non-homeserver, and all other task types remain rejected.
- Preserve the harness-lane sampler as shadow/disabled unless an independently eligible W0 proposal changes it.
- Enforce a strict allowlist of mutable fields and prevent storage/telemetry of raw prompt content, credentials, or arbitrary configuration blobs.

## Required negative coverage

Reject before any mutation: model and model-config changes; logging and test-harness changes; Gille gateway/model/tool-policy targets; deployment, auth, key, safety-gate, risk-budget, retention, and unknown targets.

## Acceptance tests

- Route-matrix regression for the current Orin boundary.
- Adapter validates version/revision/digest and accepts only the declared target schema.
- Cross-owner apply attempts leave the target untouched and emit/propose only the bound receipt surface.

## Dependencies

- Blocked by Grimnir #170, Hugin #329, and Hugin #330.
- Blocks Hugin #331.

## Non-goals

- Model roster/service configuration, Gille configuration, deploy tooling, or any protected lane.

Magnus-Gille:
## Summary

- Narrows the concrete Hugin R-exact config adapter to the sole real target, hugin-orin-macro-routing. W4.1 proposal-only prompt, harness, and tool-policy registry entries remain unchanged; this PR does not alter their receipt digest.
- Makes the macro payload a canonical unique subset, including empty, of the four reviewed homeserver classify/extract public/internal leaves, all pinned to orin and qwen2.5-coder:3b. Promotion changes the live selector and exact restore proves the baseline behavior.
- Replaces pathname-based stale takeover and lstat-then-unlink release with a persistent 0600 lock anchor and Linux kernel `flock`. The parent retains the locked open-file description for the whole synchronous mutation; close/crash releases it atomically. The implementation never renames or unlinks the canonical lock path, and missing `/usr/bin/flock` fails closed. Unsupported non-Linux developer hosts use process-local fail-fast exclusion only, without claiming cross-process safety.
- Adds adversarial Linux coverage for live contention plus SIGKILL recovery, stale-looking pathname contents that must never trigger reclaim, and deterministic injection at the former identity-lstat-to-unlink release window. The release injection runs only after mutation returns and is refused outside Vitest.
- Bounds staged documents to 64 and snapshots to 32, cleans only UUID-bound failed temp files, and emits one path-free diagnostic per missing/corrupt/contended selector failure.

## Verification

- `npm test -- --run tests/hugin-config-adapter.test.ts tests/r-exact-controller.test.ts tests/orchestrator/orin-macro-route.test.ts`: 89 passed, 3 Linux-only skipped on macOS
- `npm run build`
- `npm test`: 162 files, 2,526 passed, 3 Linux-only skipped (2,529 total)
- `git diff --check`
- Direct Linux inherited-fd probe on the Hugin host, without deploying or changing services: helper acquire `0`; competing acquire while parent retained the fd `1`; acquire after parent close `0`
- M5/mellum bounded concurrency review confirmed inherited-fd lifetime and crash-release semantics; its pathname warning was addressed by validating immediately before mutation that the acquired fd and canonical path still identify the same regular, single-link inode. Cooperating participants never rename or unlink the persistent anchor; arbitrary same-owner pathname replacement is outside that private-root threat boundary and is not claimed safe.

No production root/environment, controller/watchdog/recovery service, deployment, route change, actuator, or arming action occurred. The PR remains draft.

Closes #331

### DIFF
--- a/docs/hugin-r-exact-config-store.md
+++ b/docs/hugin-r-exact-config-store.md
@@ -1,0 +1,20 @@
+# Hugin R-exact configuration store
+
+`HUGIN_R_EXACT_CONFIG_ROOT` is the owner-installed absolute path for Hugin's
+durable R-exact configuration state. The live Orin macro selector reads only
+that store. If the variable is absent, unsafe, missing, corrupt, or contended,
+the selector returns no route; it never falls back to an in-module route map.
+
+The store is a local, mode-0700 resource. Every mutation uses an exclusive
+cross-process lock, validates the entire closed state, writes a unique fsynced
+temporary file, renames it atomically, then fsyncs the directory. Stored
+candidates and current/snapshot documents are revalidated and digest-bound on
+every read.
+
+This composition makes the macro-routing target real: staged R-exact changes
+are visible to `selectOrinMacroRoute`, persist across restart, and an exact
+recovery returns the selector to its snapshot baseline. The store does not
+install a controller/watchdog/recovery service, owner authority, journal
+backend, or an arming mechanism. Those remain a separate W0 runtime
+composition/deployment dependency; this repository ships no live root or
+environment file.
--- a/docs/hugin-r-exact-config-store.md
+++ b/docs/hugin-r-exact-config-store.md
@@ -47,13 +47,15 @@ parsing. The raw `documents` and `snapshots` objects are limited to 64 and 32
 entries respectively; those cardinalities are rejected before individual
 candidates, snapshots, or documents are deeply validated. Mutations enforce
 the same limits by retaining current and recovery-referenced documents before
-deterministic cache pruning.
+deterministic cache pruning. Snapshot references have one canonical,
+target-and-document-digest-bound form, so accepted keys are bounded as well.
 
 Within those bounds, the store validates the entire closed state, writes a
-unique fsynced temporary file, renames it atomically, then fsyncs the
-directory; a failed write removes only its own UUID-bound temporary file.
-Stored candidates and current/snapshot documents are revalidated and
-digest-bound on every read.
+unique fsynced temporary file, renames it atomically, then fsyncs the directory.
+Before creating that temporary file, it serializes and rejects any result over
+the same 256 KiB ceiling, leaving the prior readable store untouched. A failed
+write removes only its own UUID-bound temporary file. Stored candidates and
+current/snapshot documents are revalidated and digest-bound on every read.
 
 This composition makes the macro-routing target real: staged R-exact changes
 are visible to `selectOrinMacroRoute`, persist across restart, and an exact
--- a/docs/hugin-r-exact-config-store.md
+++ b/docs/hugin-r-exact-config-store.md
@@ -5,11 +5,27 @@ durable R-exact configuration state. The live Orin macro selector reads only
 that store. If the variable is absent, unsafe, missing, corrupt, or contended,
 the selector returns no route; it never falls back to an in-module route map.
 
-The store is a local, mode-0700 resource. Every mutation uses an exclusive
-cross-process lock, validates the entire closed state, writes a unique fsynced
-temporary file, renames it atomically, then fsyncs the directory. Stored
-candidates and current/snapshot documents are revalidated and digest-bound on
-every read.
+The store is a local, mode-0700 resource. Its sole concrete adapter target is
+`hugin-orin-macro-routing`; prompt, harness, and tool-policy remain
+proposal-only W4.1 concepts and have no config adapter here. The payload is a
+canonical, unique subset (including empty) of exactly four reviewed
+`homeserver` `classify|extract` `public|internal` leaves, each immutably pinned
+to `orin` / `qwen2.5-coder:3b`. A promotion therefore changes the live
+selector, and exact recovery proves the prior selector behavior.
+
+Every mutation uses an immediate exclusive cross-process lock. On Linux the
+lock records boot ID, PID start time, and its exact device/inode; only a lock
+proven stale by boot change, process absence, or PID-start mismatch is moved
+away atomically for takeover. A live, corrupt, unreadable, or contended lock
+fails closed without waiting. The selector emits at most one path-free
+diagnostic per missing/corrupt/contended reason and otherwise returns no route.
+
+The store validates the entire closed state, writes a unique fsynced temporary
+file, renames it atomically, then fsyncs the directory; a failed write removes
+only its own UUID-bound temporary file. Staged documents and snapshots are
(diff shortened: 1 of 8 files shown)


Magnus-Gille:
Root exact-head gate for `7d56cfd36ce7ab440394f679eff709245b4ce81e`: **REJECT pending real consumer composition**.

The new recovery proof is materially better: it uses the real durable store, binds the actual target scope/current candidate/recovery identity, persists `unknown` before restore, survives restart, and refuses cross-owner/stale/wrong-worker recovery.

However, the durable macro-routing target is still not the resource Hugin serves. `selectHuginMacroRoute` reads the static `defaults` object, not `HuginConfigStore`; no production `W0RuntimeGate`/journal/authority composition constructs this target. Thus `replaceExact()` can pass readback and recovery tests while changing zero live route decisions. The PR body currently says it closes #331, but the claimed strict adapter is an isolated shadow store.

Required correction: make the production selector consume a read-only, fail-closed view of the same canonical durable store (with explicit unconfigured/disarmed behavior), add a mutation→actual selector→restart→exact restore composition test, and define the production runtime/composition boundary or narrow #331/PR claims and file the missing owner-scoped follow-up. No merge, deployment, or arming.

Magnus-Gille:
Opus/root exact-head gate for `d9b17c87e3d98303dd12732ec14ebc336ff8fb0e`: **REJECT**.

The prior shadow-target defects are materially corrected: the actual selector now consumes the owner-installed store, absent/unsafe/corrupt roots fail closed, cross-process winner CAS and full state validation work, and recovery binds the durable candidate revision+digest.

Two P1s remain:

1. The file lock is `open(..., "wx")` plus `finally` unlink, with no stale-owner recovery. A real SIGKILL leaves the lock forever; restart/recovery waits ~7 s then fails permanently while the lock-free selector can keep serving the candidate. The current “crash” test throws in-process and therefore releases the lock. Require PID/boot/liveness + bounded stale-lock takeover and a real killed lock-holder recovery test.
2. Claims exceed real behavior. Three exposed targets (prompt/harness/tool-policy) have no production consumer. The sole consumed macro route schema admits only one exact route matrix/model/node, so a v1→v2 promotion cannot change any served decision; current tests cannot distinguish store consumption from a constant. Either make one bounded alternative genuinely admissible and prove a served behavior change+restore, or narrow/remove inert/unserved targets and state that this PR establishes recovery plumbing rather than autonomous configuration behavior.

P2 follow-ups: bound/prune documents and snapshots to avoid linear per-request validation growth; avoid synchronous ~7 s event-loop blocking on contention; add observable fail-closed diagnostics instead of swallowing every selector failure. No merge, deployment, or arming.

Magnus-Gille:
Follow-up pushed: 1b204a83839f20b5a3c4a4f61c95876629e41e58.

This keeps proposal-receipts unchanged to preserve W5 fixture digest compatibility, while narrowing only the concrete config adapter. It adds macro-subset selector/restore proof, immediate contention, Linux boot/PID-start/inode stale-lock proof, SIGKILL takeover and live-replacement preservation tests, bounded retention, UUID-temp cleanup, and path-free one-shot diagnostics.

Verified: focused 102 passed / 2 Linux-only skipped on macOS, TypeScript build, full 162-file suite with 2,527 passed / 2 Linux-only skipped, and git diff --check. No production or live route action occurred; draft remains intentional.

Magnus-Gille:
Follow-up quality-gate fix in 33ef858: removed both pathname TOCTOU operations (`rename`-based stale reclaim and `lstat`-then-`unlink` release). Linux now uses a persistent lock anchor plus kernel `flock`; the parent retains the locked open-file description through the mutation and releases only by closing it. Crash recovery is kernel-owned, and the canonical pathname is never renamed or unlinked. Added Linux-only adversarial tests for SIGKILL recovery, stale-looking contents that cannot trigger reclaim, and release racing a replacement pathname. Local full suite/build are green; CI is running. No deploy, route change, or arming action occurred.

Magnus-Gille:
Linux CI surfaced a test-harness bug, not a store-lock failure: command-form `flock` forks by default, so killing the wrapper left its command child alive with the inherited lock fd. Commit 3c4184d adds `--no-fork`, making SIGKILL target the actual lock-holding process. A direct Linux probe confirms acquire after that SIGKILL returns 0. Focused tests and build remain green locally; replacement CI is running.

Magnus-Gille:
Root re-review tightened the threat boundary in db474b9: removed the pre-operation `onLockAcquired` hook/test, added single-link plus canonical dev/inode validation immediately before mutation, and moved deterministic replacement injection to the exact former release boundary after mutation returns. The release hook is explicitly test-only and refused outside Vitest. We do not claim safety if an arbitrary same-owner actor mutates the pathname inside the private 0700 root; correct participants never rename/unlink the persistent anchor.

Magnus-Gille:
Final Grimnir quality gate — ACCEPT at exact head `267c12c3b59f2be8f5b6cb7b4ea468532dadb7bb` (base `d26e3d8bcd14da6bb1e87862c1f5e7de26ca2bdd`).

Evidence:
- Fresh independent Codex: ACCEPT, no grounded P0–P2; full 2,533 local tests and exact Ubuntu CI (2,560, including all Linux lock tests).
- Root Codex: ACCEPT after correcting obsolete lock docs, pre-parse/cardinality bounds, and a reproduced near-cap self-wedge.
- M5 `qwen3-coder-next-80b`: final self-wedge correction ACCEPT (`e12bcace-39aa-4078-9452-1bbb5877b4cc`); its earlier code-order misreads were grounded and withdrawn with ACCEPT (`21137e9e-c74b-4937-9bcb-1e88e0786a90`).
- Build, focused tests, full suite, diff check, and GitHub `build-test` are green.

The final store uses persistent Linux inherited-FD `flock`, no stale-path reclaim, strict 256 KiB + 64/32 bounds, canonical snapshot refs, and pre-rename write-size refusal that leaves the prior store byte-identical/readable.

Operational boundary: merge only. This PR installs no controller/watchdog/recovery service, owner authority, deployment, or arming mechanism; no production mutation or canary is performed.

Issue #324, opened 2026-07-26
from:grimnir Disable legacy hugin-daily-analysis timer


Magnus-Gille:
## Why

Grimnir issue #6 began as a response to this timer's historical five-minute Ollama timeout/thermal spike. Current host evidence on 2026-07-26 shows a different but still unresolved state: `hugin-daily-analysis.timer` is **enabled** and **active (waiting)**, with next run at 07:03 CEST; the service last ran at 07:00 CEST on 2026-07-25 and returned `Result=success` / exit 0.

The system-level observability strategy has since been superseded by the governed three-plane model in `grimnir/docs/observability-and-improvement.md`; LLM-first journal prose is not the telemetry aggregation path. The timer is no longer justified as a canary without a bounded, owner-defined success criterion.

## Requested owner action

1. Disable and stop `hugin-daily-analysis.timer` on the control host.
2. Remove/retire its unit and invocation path in Hugin, with an owner-repo regression guard.
3. Report the exact deployed revision and host-side verification (`is-enabled`, `is-active`).

After that evidence exists, Grimnir can remove the desired unit declaration from `services.json` in a separate registry change. Do not treat the current registry declaration or placement fixture as host proof.

Filed by: grimnir

Relates to Magnus-Gille/grimnir#6.

Magnus-Gille:
Operational retirement evidence: PR #325 merged at a782c6b. On huginmunin, the system timer was disabled/stopped; both system unit files were removed; daemon-reload completed; is-enabled reports not-found and is-active reports inactive. Hugin service remained active/healthy on deployed revision 6bee8c4. The first exact bound deploy of a782c6b failed closed before mutation because Grimnir services.json still declared the retired timer; Grimnir PR #160 is removing that desired declaration with this host evidence. After it merges, the exact a782c6b deploy will be rerun and verified. Reversal remains: revert #325, deploy the revert, restore the unit definitions, daemon-reload, enable/start the timer.

Magnus-Gille:
[...]

PR #342, opened 2026-08-01
Make daily analysis output Telegram-safe (#319)

### LINKED ISSUE #319: daily-analysis prompt requests markdown tables that its notification channel cannot render
`scripts/submit-daily-analysis.sh` asks for a markdown-table report under a very small output budget, and the result is delivered to a channel that cannot render tables. The output format should match the actual consumer.

### Current prompt constraints

From `scripts/submit-daily-analysis.sh`:

- `**Max output tokens:** 192`
- `Use markdown tables where appropriate.`
- `Keep the entire answer under 160 words.`
- Five requested sections (totals, per-runtime averages, cost, anomalies, quota trend)

### Why this is a bad pairing

A markdown table spends a large share of a 192-token budget on structural overhead — the header row and the `|---|---|` separator — before any data. Five sections of content do not fit alongside that overhead, so the model produces a table that is mostly scaffolding.

The primary consumer is a Telegram notification. Heimdall's `buildNotifyText` collapses newlines and truncates the body, which destroys tables specifically (filed as a separate bug against heimdall). Observed result for `20260725-050036-daily-analysis`:

```
### Response | Metric              | Value                |
|----------------------|---------------------------|| Total tasks
executed  | 1                | | Success ra
```

Even once the heimdall formatting bug is fixed, a table is the wrong shape for a chat notification.

### Suggested change

Request compact prose or `key: value` lines instead of tables — e.g. `Tasks: 1 · success 100% · avg 42s (ollama) · $0.00 · no anomalies`. This survives newline collapsing and truncation, front-loads the numbers that matter, and fits the token budget with room for the anomaly and quota sections to actually be populated.

Worth reviewing at the same time whether 192 tokens is the right budget for five requested sections, or whether the section list should be trimmed.

Note: the file header comment says the timer runs at 07:00, but observed task IDs are `20260724-050206` and `20260725-050036` (UTC). Worth confirming the comment against the installed timer unit and correcting whichever is stale.

---
Filed by: grimnir

Magnus-Gille:
Closes #319.

## Summary
- replace markdown-table guidance with compact prose or short key-value lines
- front-load totals, rates, runtime duration, cost, and the most important anomaly
- include quota trend only when evidence exists and reduce the word ceiling to 120
- pin the repository-proven 07:00 local timer contract without changing scheduling

## Verification
- focused prompt-contract test: 2 passed
- npm run build: passed
- bash syntax and shellcheck: passed
- full npm test: 157 files and 2,390 tests passed; one unrelated date-sensitive external-receipt test failed identically on canonical main
- independent Claude Opus high review: ACCEPT


Magnus-Gille:
Closed as superseded: the legacy daily journal-analysis timer, service, and submission path were retired under #324. After rebasing, this branch contains only historical documentation plus a regression test for a runtime that no longer exists; retaining that test would impose dead-path maintenance without operational benefit. No production behavior remains to change.

PR #367, opened 2026-08-12
Fix research spikes with dedicated Pi/M5 runtime

Magnus-Gille:
Closes #361
Closes #363

## Summary
- treat Pi JSON turn errors as failures even when the process exits 0
- add an explicit `Runtime: research` lane using the Pi harness against the local M5 gateway
- expose only bounded web search/fetch and exact artifact-write tools inside Bubblewrap
- keep artifact delivery and all three Munin index writes owned and verified by Hugin
- pin and verify the Pi harness during deployment

## Verification
- `npm test`: 174 files passed, 2715 tests passed, 3 skipped
- `npm run build`
- focused research/worker suites: 125 tests passed
- `bash scripts/deploy-pi.test.sh`

## M5 dogfood
- `ask` via qwen3-coder-next-80b: partial; useful only for exit-0 and artifact-boundary review, with invented implementation details discarded
- implementation subagent M5 assist via qwen3-30b-instruct: partial; all output independently tested/reviewed

PR #368, opened 2026-08-12
Fix Pi deploy path resolution

Magnus-Gille:
Follow-up to #367.

The first fail-closed production deployment installed Pi successfully under the configured user-global npm prefix, but the non-interactive SSH shell could not find it through `command -v`. Resolve and validate `npm prefix --global`, then execute Pi by its absolute installed path.

Verification:
- `shellcheck -S warning scripts/deploy-pi.sh scripts/deploy-pi.test.sh`


Magnus-Gille (commented):
Codex review gate completed at exact head 013fcb02db0982bfe100f4f6e5d9b03ab77733ec. No findings. The npm prefix is resolved remotely, required to be absolute, and the Pi binary is checked and executed by absolute path. Deploy regression and shellcheck pass; GitHub CI is green.

PR #370, opened 2026-08-13
Fix research Bubblewrap mount destinations

### LINKED ISSUE #369: Research runtime Bubblewrap mount destination is read-only
Filed by: grimnir

The first live `Runtime: research` canary on deployed revision `00fe2f09b1e6d9a780fd7bd11cd4ab943aca68a7` failed closed before Pi startup:

```
bwrap: Can't mkdir /opt/hugin-research-pi: Read-only file system
```

`buildResearchLaunch` first binds `/` read-only, then asks Bubblewrap to create `/opt/hugin-research-pi` and `/opt/hugin-research-pi-extension.mjs`. The destination parent is therefore not writable. Move runtime-only mount destinations beneath the private `/tmp` tmpfs (or explicitly create safe destinations before locking the root) and add an assertion covering mount destination ordering/paths.

Evidence: `tasks/20260813-062700-research-runtime-canary/result-structured`; no artifacts delivered, no indexes claimed.

Magnus-Gille:
Closes #369

Moves the Pi config and extension bind destinations from read-only `/opt` to the private `/tmp` tmpfs already created by the research sandbox. This fixes the first live canary failure while preserving the read-only root and exact writable artifact bindings.

Verification:
- focused research executor tests: 9 passed
- `npm run build`
- `git diff --check`

### DIFF
--- a/src/research-spike-executor.ts
+++ b/src/research-spike-executor.ts
@@ -18,6 +18,9 @@ export const RESEARCH_PI_FLAGS = [
   "web_search,fetch_content,write_artifact",
 ] as const;
 
+const SANDBOX_PI_CONFIG_DIR = "/tmp/hugin-research-pi";
+const SANDBOX_PI_EXTENSION = "/tmp/hugin-research-pi-extension.mjs";
+
 export interface ResearchSpikeRuntimeConfig {
   piCommand: string;
   bwrapCommand: string;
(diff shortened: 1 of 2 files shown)


Magnus-Gille (commented):
Codex review gate completed at exact head 130a55e53c771be53eaaca0af8ae708a2b5dfce3. No findings. Focused tests/build pass, GitHub CI is green, and the exact read-only-root + tmpfs + directory/file bind topology passed in an ephemeral Bubblewrap namespace on huginmunin.

PR #375, opened 2026-08-13
Fix Node 22 DNS lookup shape in research helpers

### LINKED ISSUE #373: Research web helpers fail under Node 22 all-address DNS lookup
Filed by: grimnir

## Production evidence

Fresh task `20260813-065454-research-runtime-canary-three` ran on accepted deployment `fa81cdf70d0dc41db466565e9c683438fd706422` using executor `research-pi-m5` and model `qwen3-coder-next-80b`.

The prior artifact-write and Bubblewrap failures are fixed: the run completed, both required artifacts were written and hash-verified on NAS, and Hugin created the three expected Munin indexes. However, every attempted `web_search` and `fetch_content` call failed with:

```text
Invalid IP address: undefined
```

The executor therefore produced ungrounded artifacts from prior knowledge. This is a stopping issue for real research spikes even though delivery now works.

## Suspected cause

`scripts/research-web-common.mjs` supplies a custom `lookup` callback to `http(s).request`. Node 22 may request `options.all === true`; in that case the callback contract expects an array of `{ address, family }` records, while the helper currently always returns scalar `address, family` arguments. The HTTP client then observes an undefined address.

Treat this as a hypothesis until reproduced and covered by a regression test.

## Acceptance criteria

- Add a regression test for the Node 22 `lookup` callback with `all: true` and preserve the scalar callback form when `all` is false.
- Keep DNS rebinding/SSRF protections intact: resolve once, reject any private result, and pin the approved public address for the request.
- Run the focused research-runtime tests and full Hugin test/build gates.
- Deploy an immutable reviewed revision and run a fresh research canary.
- The canary must successfully use public web sources, write and hash-verify both artifacts, and receive all three Hugin-owned Munin indexes.

Related: #363, #369, #371.

Magnus-Gille:
## Summary

- honor Node 22's `lookup` callback contract when HTTP(S) requests set `all: true`
- retain the scalar callback form for ordinary lookups
- keep resolve-once DNS pinning and public-address validation intact
- add regression coverage for both callback shapes

Fixes #373.

## Verification

- focused research web-helper tests passed
- `npm run build` passed
- full suite passed: 174 files, 2718 passed, 3 skipped
- `git diff --check` passed

## M5 dogfood

`qwen3-coder-next-80b` provided partial useful guidance; output was independently verified by deterministic tests.

### DIFF
--- a/scripts/research-web-common.mjs
+++ b/scripts/research-web-common.mjs
@@ -58,7 +58,9 @@ export async function requestPublicText(rawUrl, options = {}) {
         "accept-encoding": "identity",
         "user-agent": "HuginResearch/1.0 (+https://github.com/Magnus-Gille/hugin)",
       },
-      lookup: (_host, _opts, callback) => callback(null, resolved.address, resolved.family),
+      lookup: (_host, opts, callback) => opts.all
+        ? callback(null, [{ address: resolved.address, family: resolved.family }])
+        : callback(null, resolved.address, resolved.family),
     }, resolve);
     req.setTimeout(timeoutMs, () => req.destroy(new Error("Research fetch timed out")));
     req.on("error", reject);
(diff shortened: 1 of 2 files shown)


Magnus-Gille:
Codex review gate: no findings at exact head af518b92b7b28a94c2b6c7b685b8084aeb795ed6. Confirmed Node 22 all-address callback shape, scalar fallback, one DNS resolution, private-address rejection, and pinned approved address. Focused tests/build and CI pass.

